PR #2, opened 2026-07-13
fix(crud,signal): avoid UTF-8 char-boundary panic in slugify truncation

solar345:
## Problem

`base graph extract` panics on a Markdown corpus when an extracted concept label
contains a multi-byte UTF-8 character positioned so a byte falls across the 80-byte
truncation boundary in `slugify()`:

```
thread 'main' panicked at src/crud/mod.rs:75:26:
end byte index 80 is not a char boundary; it is inside 'μ' (bytes 79..81) of `...participation-measure-μ`
```

`slugify()` keeps alphanumeric chars (so Greek/accented letters survive), then does a
raw **byte** slice `&full[..80]`. When byte 80 lands inside a multi-byte char, Rust
panics and the entire extract run dies partway through the corpus.

## Fix

Back off to the nearest UTF-8 char boundary before slicing, in the two spots that
fixed-offset byte-slice arbitrary text:

- `src/crud/mod.rs` — `slugify()` (the reported crash)
- `src/signal/flow_resurface.rs` — signal-preview truncation (`&text[..80]`, same
  hazard on unicode signal text)

`&hex[..32]` (ASCII hash) and `&known_vars[..7]` (Vec slice) were reviewed and left
as-is — neither is a UTF-8 char-boundary hazard.

## Test

Added a regression test in `crud::tests` exercising a 2-byte char straddling byte 80
plus the real-world label that crashed. `cargo test slugify` passes.

### DIFF
--- a/src/crud/mod.rs
+++ b/src/crud/mod.rs
@@ -72,7 +72,13 @@ pub fn slugify(name: &str) -> String {
     if full.len() <= 80 {
         return full;
     }
-    let truncated = &full[..80];
+    // Truncate to at most 80 bytes, backing off to a UTF-8 char boundary so a
+    // multi-byte char (e.g. 'μ') straddling byte 80 doesn't panic the slice.
+    let mut cut = 80;
+    while cut > 0 && !full.is_char_boundary(cut) {
+        cut -= 1;
+    }
+    let truncated = &full[..cut];
     match truncated.rfind('-') {
         Some(pos) if pos > 20 => truncated[..pos].to_string(),
         _ => truncated.to_string(),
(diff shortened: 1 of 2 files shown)


ChristopherKahler:
Merged as 402bf3d, shipped in [v0.13.0](https://github.com/ChristopherKahler/base/releases/tag/v0.13.0). Cherry-picked so your authorship is on the commit.

Clean catch — `&full[..80]` on a byte index is a panic waiting for the first non-ASCII label, and backing off to a char boundary is the right fix. The regression test earns its place too; thanks for including the real crashing label alongside the synthetic case.

PR #4, opened 2026-08-13
Add base-help: a Claude Code coaching skill with a verified 153-pair Q&A bank

albertabrantes:
## What this is

A self-contained Claude Code skill, `/base-help`, that acts as a coach for the base CLI. Users (especially new ones) ask it anything from "how can base help me?" to "why isn't my rule injecting?", and it answers instantly from a pre-verified Q&A bank instead of re-researching every session.

## What's in the folder

- `SKILL.md`: the coaching logic. Bank-first lookup, a first-run read-only audit that writes a per-machine profile (kept in `~/.claude/`, outside this folder, so the skill stays portable), a beginner orientation built around the handoff/fork/`*end` loop, and a coach-format answer structure (answer, command, why it works, gotcha).
- `references/qa.md`: 153 Q&A pairs across 17 sections, each with a provenance tag.
- `references/commands.md`: the read-only / mutating / destructive command surfaces with flag gotchas (e.g. `-g` goes on `base rule` before the verb).

## How it was verified

Pairs were drafted against live `--help` output and the v0.11.0 source tree, merged and deduplicated, then adversarially fact-checked: a reviewer attempted to refute every mechanism claim against the source and corrected what it could refute. The bank is stamped v0.11.0; the skill treats answers as re-verify leads when the installed version differs.

## Notes for review

- The bank includes a "Known bugs (v0.11.0)" section documenting three issues found during source verification, each cited by file and line. I've filed them as separate issues so they can be triaged independently; when they're fixed, those pairs should be updated alongside.
- Folder placement (`claude/skills/`) is a proposal; happy to move it wherever fits your distribution plans (e.g. the extension/plugin mechanism).
- Everything is machine-agnostic: no personal paths, no local state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/claude/skills/base-help/README.md
+++ b/claude/skills/base-help/README.md
@@ -1,0 +1,34 @@
+# base-help: a Claude Code coaching skill for the base CLI
+
+A self-contained Claude Code skill that turns `/base-help [question]` into a coach for the `base` CLI: it answers from a pre-verified knowledge bank instantly, teaches the underlying mechanic instead of just handing over a command, and audits the local install into a per-machine profile so it can coach toward adoption gaps (zero domain rules, never-used handoffs, and so on).
+
+## What's inside
+
+| File | Role |
+| --- | --- |
+| `SKILL.md` | The skill logic: local-profile audit, bank-first answer flow, coaching format, beginner orientation |
+| `references/qa.md` | 153 question/answer pairs covering orientation, star commands, handoffs/forks, graph and memory, rules and domains, project scoping, ingestion, GraphRAG, AST navigation, hooks, relay, admin surfaces, known v0.11.0 bugs, and destructive operations. Every pair carries a provenance tag |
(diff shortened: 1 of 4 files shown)


albertabrantes:
One concrete integration idea, in case it's useful when deciding where this folder should live:

Since `base install` already wires the Claude Code hooks, it could also offer to copy this skill into place, something like:

- `base install` (or a `base skill install` subcommand) copies `claude/skills/base-help/` to `~/.claude/skills/base-help/`, making `/base-help` available immediately after install with zero manual steps.
- On `base update`, if the shipped skill folder is newer, offer to refresh `~/.claude/skills/base-help/` (the per-machine profile lives outside the folder at `~/.claude/base-help/local/`, so refreshing the skill never touches user state, and the skill re-audits automatically when the base version changes).

That would turn the placement question into a distribution answer: the repo path just becomes wherever the install step reads from (the extension/plugin mechanism would work the same way). Happy to adjust the folder location or wire any of this differently if you have a preferred shape.

PR #30, opened 2026-09-04
fix(rule): compare rule numbers as integers, not text

### LINKED ISSUE #29: `base rule add` gives every rule after the tenth the same number (string MAX on :index)
### base version

`base 0.13.15` (built from source at b1a067f)

### Host and OS

macOS (Darwin 27.0), Apple Silicon, Claude Code hooks

### What happens

Every rule added after the tenth in a domain gets the same number, 10, and lands on the same IRI `rule/<domain>/cli-10`. `base rule list` looks fine (it shows 25 rules), but the hook injection renders the cross-product of the collided rules' texts and rationales: a domain with 25 rules injected ~200 numbered lines, most of them the same text repeated with every other rule's rationale.

Cause: `next_rule_index` in `src/crud/rule.rs` runs `SELECT (MAX(?idx) ...)` over `:index` literals that are stored as plain strings, so once `"0"`..`"9"` exist the max is `"9"`, the next index is 10, and it stays 10 forever. The graph confirms it:

```
$ grep 'ontology#ruleText>' .base/graph.nq | awk '{print $1}' | sort | uniq -c | awk '$1>1'
  15 <http://ops-sys.local/ontology#rule/pool-tie-breaks/cli-10>
```

`ORDER BY ?pri` in `src/domain/query.rs` has the same string-vs-number problem (10 sorts before 2), which is only cosmetic.

### How to reproduce

```
mkdir t && cd t && git init -q && base scaffold
for i in $(seq 1 12); do base rule add --domain probe --text "rule $i"; done
grep 'rule/probe/cli-' .base/graph.nq | grep ruleText | awk '{print $1}' | sort -u | wc -l   # 11, expected 12
echo '{"prompt":"probe"}' | base hook user-prompt-submit                                     # rules 10..12 rendered as a cross-product
```

(add `prompt_keywords = ["probe"]` to the probe domain in `.base/domains.toml` and `base domain sync` first, for the last line.)

### What you expected

12 rules, 12 distinct numbers, 12 injected lines.

### Workaround, if you found one

Patch: cast to integer in both queries.

```
- "SELECT (MAX(?idx) AS ?max_idx) WHERE {{
+ "PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+  SELECT (MAX(xsd:integer(?idx)) AS ?max_idx) WHERE {{
```
in `src/crud/rule.rs`, and `ORDER BY xsd:integer(?pri)` in `src/domain/query.rs`. Verified: 12 rules get 12 numbers, and an existing collided graph is fixed by deleting the domain's rule quads and re-adding. Happy to open a PR if you want it.

marcswindle:
Closes #29.

**What changes for a user.** A domain can hold more than ten rules. Before this, every rule added after the tenth was given number 10, landed on the same IRI `rule/<domain>/cli-10`, and the hook injection rendered the cross-product of their texts and rationales (a 25-rule domain injected ~200 lines). Rules also now sort numerically in the injection (10 after 9, not after 1).

**Why.** `next_rule_index` took `MAX(?idx)` over `:index` literals stored as plain strings, so once `"0"`..`"9"` existed the maximum was `"9"` and the next index stayed 10 forever. Casting to `xsd:integer` in that query, and in the injection's `ORDER BY`, fixes both.

**Verified.** 12 rules added to a fresh domain get 12 distinct numbers; the injection lists 12 lines. Full `cargo test` passes except the two `doorbell_test` cases, which fail identically on `main` at b1a067f on this macOS machine (unix-socket accept timing), so they are unrelated.

**Existing collided graphs** are not migrated by this change; deleting the domain's `rule/<domain>/cli-*` quads and re-adding the rules repairs them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013pujoWN6FPbRgQhujseTwC

### DIFF
--- a/src/crud/rule.rs
+++ b/src/crud/rule.rs
@@ -145,7 +145,8 @@ pub fn remove(cwd: &Path, ns: &NamespaceConfig, domain_name: &str, index: u32) -
 fn next_rule_index(cwd: &Path, ns: &NamespaceConfig, domain_iri: &str) -> Result<u32> {
     let p = &ns.prefix;
     let sparql = format!(
-        "SELECT (MAX(?idx) AS ?max_idx) WHERE {{\n\
+        "PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>\n\
+         SELECT (MAX(xsd:integer(?idx)) AS ?max_idx) WHERE {{\n\
            GRAPH ?g {{\n\
              <{domain_iri}> {p}:hasRule ?rule .\n\
              ?rule {p}:index ?idx .\n\
--- a/src/domain/query.rs
+++ b/src/domain/query.rs
@@ -31,7 +31,7 @@ pub fn query_domain_from_graph(
              OPTIONAL {{ ?rule {p}:rationale ?rationale }}\n\
            }}\n\
          }}\n\
-         ORDER BY ?pri"
+         ORDER BY xsd:integer(?pri)"
     );
 
     let rules_text = match crate::store::query(store, &rules_sparql) {


ChristopherKahler:
Verified here before merging, on Windows and WSL.

Reproduced on the released 0.13.16 binary in a throwaway home: 12 rules added to one domain produced 11 distinct `cli-N` IRIs, rules 11 and 12 both numbered 10, and 6 quads on `rule/probe/cli-10`. With this commit: 12 distinct IRIs, numbered 0 through 11.

Because the branch predates #33, its own clippy run failed on lints unrelated to it. Cherry-picked onto current main the commit is clippy-clean under rustc 1.98.1 and the full suite passes.

Merging. A follow-up on main completes two things this leaves: `base rule list` still carries `ORDER BY ?pri` at `src/crud/rule.rs:85`, so the printed list still reads 0, 1, 10, 11, 2; and the inline `PREFIX xsd:` in `next_rule_index` duplicates the declaration `crud::prefixes` already prepends. A regression test lands with them.

Thanks for the report and the fix.

PR #81, opened 2026-09-07
fix(ast): every parsed extension is a file node, .baseignore dirs match on Windows (#66)

### LINKED ISSUE #66: Extensions parsed but absent from `_FILE_EXTS` lose file attribution on every platform (plus a Windows-only `.baseignore` bug)
Two bugs found on BASE 0.13.19 against a large monorepo (~17k files, mixed TypeScript /
Python / Node scripts). Both are quiet — nothing errors, the output just stops being right —
which is probably why they're still around.

**The first is not platform-specific.** I found it on Windows, but the mechanism has nothing to
do with the OS and it will affect anyone whose tree contains `.vue`, `.mjs`, `.kts`, `.mm`,
`.hpp` and friends. The second genuinely is Windows-only. Separating them because the first one
matters much more widely than where I happened to hit it.

---

### 1. Extensions in `_DISPATCH` but not in `_FILE_EXTS` lose file attribution — all platforms

`ttl_serializer.py::_FILE_EXTS` decides whether a node label counts as a **file node**:

```python
for nid, label in node_labels.items():
    if any(label.endswith(ext) for ext in _FILE_EXTS):
        file_node_ids.add(nid)
```

`.mjs` and `.ps1` are in `extractor._DISPATCH` — they're collected and parsed, and their
functions show up in the graph — but they're **not** in `_FILE_EXTS`. So their file nodes are
never identified, `_build_file_membership` never maps their entities, and `_resolve_source_file`
takes the `if not file_node_id: return fallback` branch, attributing every one of them to the
project root.

The symptom is quiet: `--contains` still finds the symbol, and the **line number is correct** —
`line_num` is computed independently of file membership. Only the file is wrong, and every
affected entity collapses onto the same project-root path. So several same-named functions in
different files become indistinguishable: you get N rows with correct line numbers and no way to
tell which file each belongs to.

```
$ base ast query --contains "readStdin"
C:\...\<repo-root>:23  fn readStdin()      # actually scripts/some-hook.mjs:95

$ base ast query --file "scripts/some-hook.mjs"
No AST entities found.
```

**Why this isn't a Windows issue.** The discriminator is
`if any(label.endswith(ext) for ext in _FILE_EXTS)` — no path separator, nothing OS-dependent.
The control is in the same run on the same machine: `.js` produced 716 attributed entities and
`.mjs` produced 0, with identical paths and separators. Both are in `_DISPATCH`; only one was in
`_FILE_EXTS`. If separators were the cause, `.js` would have failed too.

In my tree that's 600 `.mjs` + 58 `.ps1` — **25% of all collected files**. Adding just those two
to `_FILE_EXTS`:

| | before | after |
|---|---|---|
| `.mjs` attributed entities | 0 | 6,110 |
| `.ps1` attributed entities | 0 | 132 |
| entities on the project root | 6,495 (30%) | 253 (1.2%) |
| total entities | 21,979 | 21,979 (unchanged) |

After the change, all seven `readStdin()` definitions in the tree report line numbers matching
`grep -n` exactly.

The full gap, `set(_DISPATCH) - set(_FILE_EXTS)`, is 39 entries:

```
.F .F03 .F08 .F90 .F95 .bash .cc .cxx .dfm .dpk .dpr .f .f03 .f08 .f90 .f95
.gradle .groovy .hpp .inc .json .kts .lfm .lpk .lpr .luau .m .md .mdx .mjs
.mm .pas .pp .ps1 .qmd .sv .toc .v .vue
```

`.vue`, `.mm`, `.hpp`, `.cc`, `.cxx` and `.kts` look most likely to bite others. `.cjs` isn't in
`_DISPATCH` at all, so it's never parsed.

Since a hand-maintained list will drift again, two thoughts: deriving `_FILE_EXTS` from
`_DISPATCH` keys would make them impossible to desynchronise, and a debug counter for "entities
that fell back to the project root" would surface this class of problem immediately rather than
months later.

---

### 2. Directory patterns in `.baseignore` are inert on Windows — Windows-only, and much smaller

`detect.py::_is_ignored` matches a directory pattern with:

```python
if rel_str == pattern or rel_str.startswith(pattern.rstrip("/") + "/"):
```

`rel_str` is `str(path.relative_to(root))`, which on Windows is backslash-separated — so
`archive/` is compared against `archive\some\path.ts` and never matches. Verified: with
`archive/` in `.baseignore`, 139 files under it were still collected.

Only `path.match(pattern)` works there, matching the final component — so extension globs
(`*.md`) are effective and directory patterns silently aren't. Normalising `rel_str` to forward
slashes before the prefix test would fix it, and matches what `crud::normalize_path_sep` already
does on the Rust side.

---

Both verified against `origin/main` today, not just the release build. Happy to send a PR for
either if that's easier than patching them yourself.

ChristopherKahler:
Closes #66.

Reported and diagnosed by @Elevation-mindset against a ~17k-file monorepo, who measured the counts, found the cause, and offered a patch. Both defects reproduced here before anything was changed.

## 1. Every parsed extension is a file node

`scripts/ast/extractor.py` parsed **66** extensions. `scripts/ast/ttl_serializer.py` recognised **27** of them as files. The 39 in the gap were parsed, reached the graph, and had every one of their entities attributed to the **app root** rather than the file they came from, because `_build_file_membership` never saw a file node for them and `_resolve_source_file` took its fallback. Line numbers stayed correct, so `base ast query` answered with a real line in the wrong file, and `--file src/thing.mjs` found nothing.

The measured gap, recomputed on `972996a`:

```
.F .F03 .F08 .F90 .F95 .f .f03 .f08 .f90 .f95   -> fortran   (10)
.pas .pp .dpr .dpk .lpr .inc                     -> pascal     (6)
.md .mdx .qmd                                    -> markdown   (3)
.cc .cxx .hpp                                    -> cpp        (3)
.m .mm / .gradle .groovy / .v .sv / .luau .toc / .mjs .vue      (10)
.dfm .lfm .lpk .kts .ps1 .json .bash                            (7)
```

`set(_FILE_EXTS) - set(_DISPATCH)` was empty, so the fix only ever *adds*: nothing an existing map treats as a file stops being one.

**The fix.** In a whole-app extraction `onto_ast.py` already builds `file_map` keyed by every id form a file node can carry, so membership in it **is** the definition of a file node — exact, and needing no extension list at all. That is now the primary test (`_identify_file_nodes`, `ttl_serializer.py:139`). The extension set that covers single-file mode, where there is no `file_map`, is derived from the parser table (`_FILE_EXTS = frozenset(_DISPATCH)`) rather than typed out beside it, so the two cannot drift again. There is deliberately no `try/except` around that import: a degraded empty set would attribute *every* entity to the app root, silently, which is the failure this fix exists to remove.

### Before / after, on two real trees

Measured on this machine, on Windows, with the shipped code path:

| tree | files | entities | on app root before | after |
|---|---|---|---|---|
| a hub app (11 `.mjs`, 6 `.md`, 1 `.json`) | 113 | 2,807 | **293 (10.4%)** | **89 (3.2%)** |
| a mod toolchain (`.ps1`, markdown-heavy) | 108 | 2,027 | **1,340 (66.1%)** | **1,154 (56.9%)** |

File nodes recognised went 96 → 113 and 79 → 108. The two predicates (exact `file_map` membership vs the derived extension set) were compared on both trees and **agree exactly, with an empty disagreement set in both directions**.

The residue is a **separate, pre-existing defect**, not something this PR introduces: `_build_file_membership` walks containment exactly one hop from a file node, so a nested markdown heading or an `inherits` chain still orphans. It is reproducible on `972996a` with `.cpp`, an extension that was always a file node — the member function of a struct is attributed to the app root there too. Filed as #82; the new counter below is what makes it countable.

### Side effect worth naming

`.hpp` was already in `_STRUCT_LANGUAGES` but never in `_FILE_EXTS`, so a struct in a C++ header could not be typed as a struct. It can be now, so a map diff will show those retyped.

## 2. `.baseignore` directory patterns on Windows

`_is_ignored` compared a forward-slashed pattern against `str(path.relative_to(root))`, which is backslash-separated on Windows, so `archive/` never matched `archive\deep\thing.ts` and 139 files under an ignored directory were collected anyway. `rel.as_posix()` normalises both sides, matching what `crud::normalize_path_sep` already does on the Rust side.

Narrower than first described: only patterns **carrying a slash** were affected. A bare `archive` already matched through `rel.parts`, and a glob like `*.log` already matched through `path.match`. POSIX behaviour is unchanged and is pinned by its own test.

Proven on Windows (`sys.platform = win32`, `os.sep = '\'`), same fixture, both script generations:

```
0.14.1   files walked=2  archive/ leaked=True   control a.mjs present=True
branch   files walked=1  archive/ leaked=False  control a.mjs present=True
```

## 3. The extractor says when it could not place an entity

A run that falls back to the app root now reports how many, always rather than only in devmode; a run with nothing to report prints nothing new. This is the tripwire that would have caught the drift above.

`base sync --ast --yes` piped the extractor's stderr so a **failed** build could explain itself at the next session start, and on a **successful** build captured and discarded it, including every "skipped this file" line. `echo_extractor_notices` (`src/hook/automap.rs:1121`) forwards the `# ` notices on success; the failure path and `.last-error` are untouched, and the non-`--yes` path is untouched.

Scope, precisely: this makes the notices visible on a **foreground** `--yes` sync. **A background refresh still shows nobody anything** — `spawn_sync` hands its child `Stdio::null()` and the git hook redirects to `/dev/null`, both deliberately. Reaching the hook path means persisting the notices for session start to read, the way `.last-error` already works for failures; filed as #89 rather than bolted on here.

## 4. `.cjs`

Was in none of the three extension lists, so it was never parsed at all. One line.

## Tests

`tests/ast_extension_registry.rs` — four tests riding `cargo test`, no tree-sitter grammar required, so they add no CI dependencies:

- the dispatch/file-node drift assertion, in both directions, plus `_FILE_EXTS == frozenset(_DISPATCH)`
- `.cjs` is parsed and is a file node
- `.baseignore` directory patterns against a `PureWindowsPath`, so the Windows behaviour is measured on a Linux runner, with the four already-working rows pinned against regression
- POSIX behaviour unchanged

`tests/ast_extractor_notices_test.rs` — the notice filter, and an end-to-end `sync --ast --yes` against a stubbed extractor asserting the count reaches the caller on success while non-notice lines do not.

The extraction rows that do need grammars (per-extension attribution, the counter, the CLI read-back, the Windows `.baseignore` run) live in `verification/base-0.14.2/ast_attribution_0142.sh`, which refuses to report any row unless it can show an md5 difference and a branch-only symbol between the two script generations. **16 PASS / 0 FAIL / 1 SKIP** locally; the remaining SKIP is the `.baseignore` row's RED half, which can only go red on Windows and is run there separately.

## Upgrading an existing map

Attribution lives in `ast.ttl`, so a map built before this release keeps its old paths until it is rebuilt. `base sync --ast --target <app>` does that in seconds: the per-file extraction cache stores `{nodes, edges}` and is still valid, only the serialisation step re-runs, and every extractor already emitted a file node for every parsed file — it simply was not recognised as one. The Stop hook does the same on its own the next time you work in that app, unless the tree is over the 0.14.1 size fuse, in which case run the command yourself. Nothing re-indexes automatically.

**Also filed while building this:** #82 (one-hop containment, the residue above), #83 (`LANG_MAP` is a third extension list that has drifted the same way), #84 (`.vue` is in the dispatch table but an SFC is unparsable), #85 (nothing in CI runs the Python AST tests, which is how this drift survived), #89 (a background refresh's notices reach nobody).

https://claude.ai/code/session_01E2UH6ioGZoDta1ts6Aigc3


ChristopherKahler (commented):
Second-builder review (0.14.2 merge rule, line 6) - merlin (Fable), relayed by the orchestrator auk.

Verdict on f58c8f3: **APPROVE**.

Read: the full delta 003b3b7..f58c8f3 (detect.py, ttl_serializer.py, extractor.py, onto_ast.py, src/hook/automap.rs echo_extractor_notices, src/cli.rs --yes success arm, tests/ast_extension_registry.rs, tests/ast_extractor_notices_test.rs, the harness, the changelog). Code delta correct.

One blocker was raised on 7a9b0a6 and cleared by the reword in f58c8f3: the echoed extractor notices cannot reach anyone on the Stop-hook path (spawn_sync sends the child's stdout+stderr to Stdio::null, automap.rs) or the git-hook path (ast_repo.rs redirects to /dev/null), and five places claimed they survived there. On f58c8f3 every claim states foreground-only visibility and names both redirects; no residual claim remains (grep). Persisting notices for session start is filed as #89.

CI on f58c8f3: test, clippy, release-invocations success (API).

Issue #87, opened 2026-09-07
10 graph writers still write unlocked (bulk/maintenance sites)


ChristopherKahler:
The graph write lock added in #72/#73/#74 covers the eleven writers on the hot path. Ten more still do an unlocked load → mutate → write, so two of *them* racing still lose a write silently. This issue carries the list and each site's shape, so the remaining work is a decision rather than a rediscovery.

Background: `graph.nq` is a whole-file read-modify-write. Without serialization the writer that renames second overwrites the first, and both report success — measured on 0.14.1 as eight concurrent `fork create` leaving three of eight rows on Windows and two of eight on Linux, with no non-zero exit anywhere.

## Migrated already (for reference)

Three shared seams covering 40 callers, plus eight that do their own load:

`crud::load_and_mutate` (37 callers, incl. the relay ping writer), `crud::load_read_then_mutate` (3), `crud::handoff::mutate_file`, `crud::repair_edges`, `crud::supersede`, `crud::note` ×3, `graph::purge_stale`, `hook::post_tool_use`, `protocol::reconcile`.

## Still unlocked — one-shot, bulk and maintenance writers

| site | what it does |
|---|---|
| `src/extract/mod.rs:188` | markdown extraction, whole-file rebuild |
| `src/extract/paul_toml.rs:363` | paul.toml ingest, whole-file rebuild |
| `src/extension/ingest.rs:130` | extension ingest |
| `src/standards/sync.rs:322` | standards sync |
| `src/domain/sync.rs:200` | domain sync |
| `src/apply_ops.rs:479` | remote op application |
| `src/doctor.rs:793` | `doctor.repair` |
| `src/graph.rs:45` | `graph.compact` |
| `src/migrate.rs:381` | migration |
| `src/store.rs:74` | `migrate_trig_to_nq` |

Each loads a graph, rebuilds or repairs most of it, and writes it back. They are vulnerable — a `base sync` racing a hook write loses one of them — but they are not the shape of the filed issues and none is on the per-tool-call path.

Wrapping them is the same mechanical change as the eleven already done. The one open question is the wait bound: `graph.compact` and `doctor.repair` hold the lock for the length of a whole-file rebuild, a much longer critical section than a single UPDATE, so the current 10s bound wants revisiting for those two specifically.

## Out of scope: the dashboard

[...]

Issue #86, opened 2026-09-07
relay ping: two pings in the same millisecond collide on one slug, and the earlier one is silently replaced


ChristopherKahler:
Two `base relay ping` commands that land in the same millisecond get the same slug, so the second silently replaces the first: same inbox alert filename, same graph IRI. The receiver never sees the earlier ping and nothing reports a problem.

Found while building the concurrency harness for #72/#73/#74. It is not caused by that work and reproduces on 0.14.1.

## Repro

Fake home, fake workspace, eight pings fired concurrently at one target:

```sh
export BASE_HOME=/tmp/pingcon/home
export CLAUDE_CODE_SESSION_ID=00000000-0000-0000-0000-0000000pcon
base relay register --as pcon
for i in $(seq 1 8); do ( base relay ping --to pcon --from s$i --msg "m$i" ) & done; wait
```

Measured on `base 0.14.1` (linux asset from release 384286377, md5 `da7950677bce8059247e7dd858f9016b`) and identically on the branch for #72:

| | |
|---|---|
| commands run | 8 |
| `changes.jsonl` ping INSERTs | **8** — every write landed |
| inbox alert files | **2** |
| distinct `ops:ping/…` IRIs in the graph | **2** |
| the two slugs | `ping-1788816450164`, `ping-1788816450165` — one millisecond apart |

Sequentially the same eight pings produce 8 distinct slugs and 8 rows, because the gaps are 70-500 ms.

## Why

The slug is `ping-<epoch millis>`. `src/relay/task_inbox.rs` builds the IRI with `crud::build_iri(ns, "ping", &task.slug)` and names the inbox alert file after the same slug, so two pings inside one millisecond collide in both places at once. The graph write is an `INSERT DATA` at a fixed IRI preceded by a `DELETE` of that IRI, so the collision is a replace, not a duplicate — which is why it leaves no trace.

This is why the ping half of the #72 harness row could not simply assert "8 pings, 8 rows": the slug scheme cannot produce eight distinct slugs in two milliseconds. That row now asserts the invariant instead — every ping that wrote an inbox alert still has its graph row.

## What it costs in practice

Bursts are the normal case, not the edge case. An orchestrator fanning a message out to several builders, or several sessions replying at once, is exactly the shape that collides. The lost ping is indistinguishable from one that was never sent: the sender sees `Ping → target: …` and a success exit.

## Suggested shape

[...]

PR #88, opened 2026-09-07
fix(registry): serialize graph writes, and make archive/create say what they did (#71 #72 #73 #74)

### LINKED ISSUE #71: base handoff create does not archive the prior open handoff, though its help says it does
### What happens

`base handoff create` documents itself as archiving the prior open handoff for the project, in both the group help and the subcommand help:

    Commands:
      create   Register a handoff doc (archives any prior open handoff for the project)

It does not. Registering a second handoff for the same project leaves the earlier one `open`, so both resurface at session start.

### Reproduced

Two handoffs for project `base`, the second registered at 15:07:23 while the first had been open since 10:49:16:

    $ base handoff create --project base --doc .../2026-09-07-plover-base.md
    Handoff for 'base' registered (slug: 2026-09-07-plover-base)

    $ base handoff list
    | 2026-09-07-plover-base | base | open | 2026-09-07T15:07:23-05:00 |
    | 2026-09-07-auk-base    | base | open | 2026-09-07T10:49:16-05:00 |

The older one is still `open` after the create that should have archived it.

### Why it matters

A resume doc that outlives its round keeps resurfacing, and the count grows with every session that ends properly. `base doctor` is already carrying an advisory about the number of open handoffs, which is the same defect seen from the other end. Someone closing a round has to archive by hand and has no reason to think they need to, because the help says the tool did it.

### Either fix is fine, but pick one

Make `create` archive the prior open handoff for that project as documented, or change both help strings to say it does not. The current state is worse than either, because the wrong one is written down.

### Found

Closing out the 0.14.1 round. An orchestrator had deliberately sequenced its own registration last to work around the documented archiving, and the archiving never happened.

### LINKED ISSUE #72: fork archive from a workspace prints 'archived' and changes nothing when the fork lives in the global tier (#55's shape on the fork/handoff registries)
0.14.1 (`base.exe`, Windows install, cwd `C:/Users/Chris` = the workspace tier), fork `base-injection-scope` registered in the GLOBAL tier (`~/.base-gbl/forks/`):

```text
> base fork archive base-injection-scope
Fork 'base-injection-scope' archived
> base fork list | grep injection-scope
| base-injection-scope | base | open | C:/Users/Chris/.base-gbl/forks/base-injection-scope.md |
```n
The workspace-tier call printed success and changed nothing; `base fork -g archive base-injection-scope` was what archived it (`| base-injection-scope | base | archived |`). Same shape as #55 on the fork registry: a write that acts on the tier you stand in must say what changed and exit non-zero when nothing did, naming the tier it searched and the one it did not; `fork create/archive/snooze` and `handoff create/archive/snooze` were not covered by PR #61's tier seam. Red-first: from a workspace, archive a fork registered only in the global tier and assert rc != 0 and a sentence naming both tiers.

Hit by merlin during the 0.14.1 *end sweep (auk's instruction to file).

https://claude.ai/code/session_01Xnh8a9HJxoUvuFL21DPhkv

### LINKED ISSUE #73: base fork create then archive in one invocation reports success and leaves no row at all
`base fork create` followed by `base fork archive` reported success on both and left **no row in any state** — the slug absent from `base fork list` entirely, neither `open` nor `archived`.

## Repro

Both commands in **one shell invocation**, back to back, with no read between them:

```
base fork create --project base --doc "C:/Users/Chris/.base-gbl/forks/base-supersedes-drift.md"
base fork archive base-supersedes-drift
```

```
Fork 'base-supersedes-drift' registered for 'base'
Fork 'base-supersedes-drift' archived
```

Then:

```
base fork list | grep supersedes-drift     -> nothing
base fork list | grep drift                -> nothing
base fork list | grep archived | wc -l     -> 48
```

Forty-eight archived rows were listed and none of them was the one just archived. A row archived earlier in the same sweep, `base-f25-ingest-idempotence`, was present and correct in that same listing, so the listing itself was working.

## What separates the failing run from the passing one

Running the same two commands as **separate invocations** produced the correct result:

```
base fork create ...        -> | base-supersedes-drift | base | open     | ... |
base fork archive ...       -> | base-supersedes-drift | base | archived | ... |
```

`create` alone produces a listable `open` row, so the row does exist between the two commands. The difference is only how closely the two writes follow each other.

## Confidence

Observed once, did not reproduce on retry. That makes it timing-shaped, and **I am not asserting the mechanism** — I have not read the write path. A same-slug read-modify-write where the second command loads a store snapshot taken before the first one landed would fit, and so would several other things.

## Why it is worth an issue rather than a shrug

Both commands printed success. Nothing in the output distinguished the run that lost the row from the run that kept it, and the only reason it was caught is that the `*end` verification line reads the registry back. A registry write that reports success and silently leaves nothing is the same class as the stale-instrument and false-pass defects this release round spent the day removing: the failure is invisible at the point it happens.

Found during the 0.14.1 `*end` sweep, 2026-09-07, on the released binary.

### LINKED ISSUE #74: fork archive leaves no row: after 'base fork archive <slug>' printed archived, the slug is absent from 'base fork list' (neither open nor archived)
Observed 2026-09-07 16:40-16:55 on 0.14.1 (Windows install E3E3C076, cwd C:/Users/Chris, forks registered in the global tier).

1. 'base fork list' showed base-prompt-time-traversal, base-domain-normalisation and base-domain-normalisation-verification as open.
2. 'base fork archive <slug>' x3 printed 'Fork ... archived'.
3. 'base fork list' afterwards lists NONE of the three: no open row, no archived row. Other archived forks (base-tier-cli-cluster, base-0.14.1-hooks-dashboard, base-injection-scope) do list as archived.
4. shrike saw the same shape once for base-supersedes-drift (create then archive in one shell invocation: no row at all), not reproduced on retry.
5. The three docs had their frontmatter set to 'status: shipped' just before the archive; if the listing filters on the doc status rather than the registry, an unknown status hides the row. Untested.

Related: #72 (workspace-tier archive touches nothing for a global-tier fork) and #71 (handoff create archives per tier, not per project). Expected: an archived fork lists as archived from either tier; a status the registry does not know is reported, not hidden.

Milestone 0.14.2. Filed by auk (orchestrator) at the 0.14.1 close-out.

### PR DESCRIPTION
**Pull request by ChristopherKahler**
Closes #71. Closes #72. Closes #73. Closes #74.

Three of the four issues turned out to be one defect. `graph.nq` is a whole-file read-modify-write with nothing serializing the writers, so two `base` processes overlapping destroy each other's changes and **both report success**. The fourth (#71) is a separate, real reporting gap.

The fork doc framed this as a Windows `fs::rename` failure. It is neither Windows-specific nor a failure: Linux 0.14.1 loses harder than Windows, and across ~150 concurrent writes not one command exited non-zero.

## The four repros, before and after

Control: `base 0.14.1`, linux asset from release 384286377, md5 `da7950677bce8059247e7dd858f9016b`.
Branch: md5 `092630efe8e04540f840835179a45d6a`.
Harness: `verification/base-0.14.2/registry_0142.sh`, `BASE_BIN`-driven, fake `BASE_HOME`, isolation from the real graphs asserted in-run.

**#74 — archive prints `archived`, the row stays `open`**

```sh
for i in $(seq 1 8); do ( base fork archive "a-$i" ) & done; wait
```
| | 0.14.1 | branch |
|---|---|---|
| printed | 8 × `Fork 'a-N' archived` | same, each naming its tier |
| actually archived | **2 of 8** | **8 of 8** |

**#73 — the row is gone in every state**

50 × `fork create` + `fork archive` with one concurrent `base` writer on the same graph:

| | 0.14.1 | branch |
|---|---|---|
| rows absent | **16 of 50** | **0** |
| present but not archived | **8 of 50** | **0** |
| non-zero exits | 0 | 0 |

Sequentially, without another writer, 0.14.1 is 50/50 clean — the defect needs a second writer, which on a real machine is every hook.

**#72 — archive of a slug nothing holds**

```sh
base fork archive shrike-no-such-slug-xyz
```
```
0.14.1:  rc=0   Fork 'shrike-no-such-slug-xyz' archived
branch:  rc=1   Failed: no fork 'shrike-no-such-slug-xyz' in either tier — nothing was archived. Searched:
                  <ws>/.base/graph.nq (workspace tier)
                  <home>/.base-gbl/.base/graph.nq (global tier)
```

The cross-tier half of #72 does **not** reproduce: a global-tier fork archives correctly from a workspace cwd with no `-g`, first try, on 0.14.1. Entity IRIs are tier-invariant and `archive` binds `GRAPH ?g` as a variable, so the graph IRI never entered the match. What made it look like a tier bug was the lost write plus the unconditional success print. That leg is kept as a green-both-sides regression pin.

**#71 — create archives the prior handoff silently**

```sh
base handoff create --project pj --doc HANDOFF-A.md
base handoff create --project pj --doc HANDOFF-B.md
```
```
0.14.1:  Handoff for 'pj' registered (slug: HANDOFF-B)
branch:  Handoff for 'pj' registered (slug: HANDOFF-B)
         archived prior open handoff: HANDOFF-A (workspace tier)
```
and with an open handoff in the other tier:
```
branch:  Handoff for 'pj' registered (slug: HANDOFF-W)
         no prior open handoff in this tier
         global tier also holds an open handoff for 'pj': HANDOFF-G — archive it with: base handoff -g archive HANDOFF-G
```
The tier scoping is correct and stays (#61: a write acts on the tier you stand in). Both help strings now say "in this tier" instead of promising a project-wide archive.

## The decisive row

R13 reproduces the production burst: on 2026-09-07 the global graph took five writes between 15:07:14 and 15:07:15 — an `ops:Ping` INSERT and four archive UPDATEs. The interleaver was `relay ping`, not a registry command.

8 concurrent `fork -g archive` interleaved with 8 `relay ping`:

| | 0.14.1 | branch |
|---|---|---|
| archives that survived | **2 of 8** | **8 of 8** |
| ping rows kept, vs pings that alerted | **0 for 6** | **6 for 6** |

A registry-only lock would have left this standing.

## What changed

`store::with_graph_lock` / `lock_graph`: an exclusive lock file beside `graph.nq` so the tiers lock independently, bounded 10s wait, a loud non-zero error naming the lock path on timeout, stale reap by mtime on the same 60s rule the existing abandoned-temp sweep uses, released on `Drop`. **Re-entrant per path** — an outer writer calling a helper that locks the same graph must not self-deadlock. **Readers never take it**: `fs::rename` stays atomic and hooks read on every tool call.

The critical section spans load → update → write_back. A store loaded before the lock is exactly the stale snapshot that loses the write, so `crud::workspace_graph_path` and `crud::lock_and_load` split path resolution from loading.

Eleven writers migrated: three shared seams covering 40 callers (`crud::load_and_mutate` — 37, including the relay ping writer — `load_read_then_mutate`, `handoff::mutate_file`) plus `repair_edges`, `supersede`, `note` ×3, `graph::purge_stale`, `hook::post_tool_use`, `protocol::reconcile`.

`archive`/`snooze` now ASK each tier file whether it holds the slug before mutating, on the store already loaded inside the lock, and return the tiers they changed. `all_tier_files` deduplicates by canonical path, so `-g` stops running the UPDATE twice on one file (visible in production as two identical archive lines at the same second).

## Cost

Ten runs each, medians, on a copy of a real 16.5 MB graph:

| | 0.14.1 | branch | delta |
|---|---|---|---|
| `fork create` | 1135 ms | 1154 ms | +19 ms |
| `fork archive` | 1095 ms | 1017 ms | −78 ms |
| `hook post-tool-use` (every tool call) | 1029 ms | 1010 ms | −19 ms |

All inside run-to-run spread; the control's own `create` ranged 1072-1455 ms.

## Verification

- Harness: **22 pass / 0 fail** on the branch, **4 pass / 13 fail** on the 0.14.1 control, both provenance-gated by md5 and a branch-only marker (the harness aborts on a mismatch).
- `tests/registry_lock_test.rs`: 12 tests, one per lock. Mutating the lock out (`with_graph_lock` → `f()`) fails exactly the 3 lock tests and passes the other 9, so the tests can actually fail.
- `cargo test --all-targets` green; `cargo clippy --all-targets -- -D warnings` clean.
- `tests/prove_walk_hook.sh` PASS on the branch binary.
- CLI reference regenerated (`BASE_REGEN_DOCS=1`) for the two changed help strings.

## Filed separately, not fixed here

- **#86** — `relay ping` slugs are `ping-<epoch millis>`; two pings in one millisecond collide on one slug and the earlier is silently replaced in both the inbox and the graph. Pre-existing, unrelated to the lock.
- **#87** — the 27 writers still unlocked: the dashboard's 17 (blocked on its long-lived cached store) and 10 one-shot/bulk sites, each with file:line and shape.

https://claude.ai/code/session_01Cmtz7uD6taWbZQEhYmeEYU


ChristopherKahler (commented):
**APPROVE** — second-builder review (0.14.2 merge rule, line 6), on `8c1af26`.

_Posted as a comment: GitHub refuses a self-authored APPROVE state, so the verdict is the first line of the body — the same form as the PR 79 and PR 94 reviews._

**Rebase verified independently before reviewing**, not taken on report: `merge-base(origin/main, 8c1af26) == e62c91b == origin/main` (clean rebase), `git diff 3fc7250..8c1af26 -- src/ tests/ Cargo.toml Cargo.lock` is **0 lines**, and the four tree/blob hashes are identical — `src b7567ce4eb0d`, `tests c9e824acaf5b`, `Cargo.toml 1fdebad12501`, `Cargo.lock 8bb7ec123867`. So the code reviewed below is byte-identical to the code at this sha.

---

**1. The lock seam closes the no-op writers rather than papering them.** The distinction is *lock before load*, and it holds at every site I checked: `purge_stale` takes `lock_graph` on the line before `load_graph` (`src/graph.rs`), `post_tool_use::handle` does the same before its `load_graph`, and `locked_update` puts the load inside the closure — `load_or_empty` then `update_and_write`, both inside `with_graph_lock`. A lock taken *after* the load would serialise the writes while leaving each one computed from a stale read, which is the papering failure; that is not what this does. The `post_tool_use` comment names why it matters most there — it fires on every tool call, so it is the writer most likely to be mid-dump when another process renames over the same graph.

**2. `with_graph_lock` is used consistently and cannot deadlock.** Four mechanisms, all present:

- **Re-entry** is handled by the `HELD_LOCKS` thread-local: a nested `lock_graph` on a path this thread already holds returns a `reentrant: true` guard without touching the file, and `Drop` returns early for it. Without this, a locked writer calling another locked writer would self-deadlock against its own lock file.
- **Bounded wait**: `LOCK_WAIT` 10s with 5ms→100ms exponential backoff, then a `bail!` that states plainly "Nothing was written." A caller blocks for a bounded time and gets an honest error, not a hang.
- **Stale reaping is pid-gated, not age-gated.** `reap_stale_lock` requires both an mtime older than `LOCK_STALE` *and* a dead holder pid, and the comment records the reason: `graph compact` and `doctor --repair` can legitimately hold the lock past the age threshold, and reaping one mid-write is how the lock becomes the bug it prevents. An unreadable pid counts as alive — the safe default.
- **`Drop` is ABA-safe.** It removes the file only when `lock_pid == std::process::id()`. Deleting by path alone would, after a stale reap and a fresh acquisition by another process, delete *that* holder's lock and put two writers on one graph. Credited in-comment to plover's earlier review; worth keeping visible.

Consistency is enforced mechanically rather than by convention — see 3.

**3. `registry_0142.sh` cannot pass having visited zero.** Two independent guards:

- The isolation row counts `checked` and, at zero, calls `bad "isolation proved NOTHING — no operator graph was found at any known path"`. The comment above it records the exact prior false green this fixes: WSL-only path spellings made every Windows run take the `[ -f ] || continue` branch on both entries and print PASS having opened nothing, in the one row whose whole job is proving nothing leaked. Both spellings are now listed.
- `EXPECT_LOCKED` compares a branch-only marker — `grep -ac 'waiting for the graph lock'` against the binary — and **exits 2** on mismatch: "That is the wrong binary — refusing to report numbers from it." That is provenance enforced by the harness rather than asserted in prose.

The same discipline is in `tests/lock_tripwire_test.rs`, which is a source scan carrying two law-23 asserts of its own: `!files.is_empty()` ("the tripwire measured nothing") and `checked > 0` ("it would pass on an empty scan"). It requires every function containing a write call to contain a lock token, with an allow-list that prints each entry and its reason.

**4. No R16 leg can go green without the fix.** This was the question I most wanted to answer, and the harness answers it three ways: against a control binary the leg prints `R16 skipped: control binary has no lock` — a **skip, not a pass**; against a wrong binary `EXPECT_LOCKED` aborts before any leg runs; and if the leg cannot obtain a live OS-level pid it calls `bad "R16 could not obtain a live os-level pid — leg DID NOT RUN"`, so a could-not-run is a **FAIL, not a silent skip**. That is the separation between *failed* and *could-not-run* that a gate needs.

---

**Nothing blocking.** One thing I checked and cleared rather than flagged: `open_handoff_in` (`src/crud/handoff.rs`) calls `load_or_empty` with no lock, which reads like a gap. It is a read-only `SELECT` returning `Option<String>` and performs no write, so it is correctly outside the seam.

PR #90, opened 2026-09-07
feat(install): one-line installers, no Rust toolchain, above the fold

ChristopherKahler:
feat(install): one-line installers, no Rust toolchain, above the fold

The README's front door was a source build. On Windows that additionally meant
the MSVC developer environment and LLVM/libclang. v0.14.1's four release assets
were downloaded 1, 2, 4 and 6 times; across all 59 releases the assets have
1,068 downloads total. The binaries are built every release and nobody finds
them.

Adds install.sh and install.ps1. Neither invents anything: they detect the
platform, pull the matching asset from the latest release, unpack it, and hand
off to `base install`, which already copies the binary to ~/.local/bin, writes
~/.base-gbl/, installs the AST scripts and wires the hooks.

The one-liner now sits directly under the tagline, before the prose. Quick start
leads with it for both platforms; the cargo path moves into a "Build from
source" details block, unchanged.

Two behaviours that only show up on a clean machine, both found by running the
installer end to end into an isolated HOME:

- `base install` resolves scripts/ast by trying two paths relative to the binary
  and then the working directory (install.rs:931-948). A release archive
  satisfies only the third, so the installer runs the binary from inside the
  unpacked directory. Without the cd it prints "scripts/ast/ not found near
  binary - skipped" and the AST extractor never lands.
- Hooks are only written into an existing ~/.claude (install.rs:791). Installing
  base before Claude Code leaves an inert binary that cannot self-heal, because
  the repair path is the session-start hook that was never wired. The installer
  detects this and says to rerun `base install`. It tests for settings.json, not
  the directory, because `base install` creates ~/.claude itself for the bundled
  skill and the directory therefore always exists afterwards.

Also handled: linux-aarch64 has no published asset, so it fails with that
sentence instead of a 404; a non-archive download is caught before tar sees it;
curl or wget, whichever is present; BASE_VERSION pins a tag; arguments pass
through to `base install`.

Verified on Linux end to end into an isolated HOME: exit 0, binary at
$HOME/.local/bin/base reporting 0.14.1, .base-gbl/scripts/ast/extractor.py
present, PATH notice correct, real home untouched. dash and bash both parse it.
install.ps1 is unrun.

The install URLs point at raw.githubusercontent.com so the README is never
broken. Swapping them to basemode.ai/install.sh is a one-line change once the
site serves the file.

https://claude.ai/code/session_0181pwiXBWbMcJCnHXErV5RZ


ChristopherKahler:
**Orchestrator re-gate (auk) on `eeaf4f1`.** My earlier G2 was on `99c0d11`; two blockers landed after it and the branch was rebased onto `de301d5`, so that verdict does not stand and this replaces it. **13 PASS / 0 FAIL on Windows, 8 PASS / 0 FAIL on Linux, all from my own commands.**

**Windows** (`C:\auk-pr90-regate`, `install.ps1` from the head, md5 `31728475`), three legs:

| leg | result |
|---|---|
| **CLEAN** machine, no `~/.claude` | rc 0, `base.exe` written, runnable by name from `cmd.exe` → `base 0.14.1` |
| **UPGRADE** over a stale `base.exe` — *the leg my first run did not have* | rc 0; a deliberately-wrong 4098-byte `base.exe` seeded at md5 `51A77D34` came back at **`E3E3C076`** (the real 25,297,408-byte 0.14.1 binary), the upgraded file runs and reports `base 0.14.1`, and the script *says which of the three things it did*: `refreshed base.exe`. This is the blocker merlin found — before the fix an existing `base.exe` was left untouched and the update was a silent no-op |
| **REPAIR**, Claude Code appears afterwards | `base install` wires all five hooks |

**Caller-scope leak check** (the second blocker), measured in a child PowerShell session *after* a full run rather than argued: `EAP=Continue`, `DIE=absent`, `TMP=absent`, shell alive. `SAY` resolves to an unrelated `say.cmd` already on this machine's PATH — an `Application`, not a leaked function.

**Live-machine tripwire**, content hash on both sides, never mtime: the operator's real `~/.claude/settings.json` (`45A5F7BB…`) and real `~/.local/bin/base.exe` (`E3E3C076…`) are **byte-unchanged** across all three legs.

**Linux** (`~/.cache/auk/pr90-linux2-eeaf4f1/`): rc 0, **8 PASS / 0 FAIL** — binary present, runnable by name on a PATH holding only the fresh install, `scripts/ast/extractor.py` landed, the no-`~/.claude` warning printed, hooks correctly not wired, install reached `Seed system rules ✓`, and the repair leg wires all five hooks. Live global tier and `settings.json` byte-unchanged.

**One correction to my own instrument, since it appears in my earlier numbers.** My Linux leg printed a `doctor` line labelled "on the fresh install" that read **UNHEALTHY** on this run and **HEALTHY** twenty minutes earlier. Neither was about this PR: `HOME` isolates the *home* tier only, the *workspace* tier still resolves upward from cwd, and my run's cwd was `/mnt/c/Users/Chris` — so the line was reporting the **operator's live 66,509-line workspace graph** and its path-trigger and backup advisories, which changed between the two runs. Re-measured with cwd inside the fake root: **HEALTHY**, rc 0. With cwd somewhere with no workspace tier at all: `no workspace here: searched the global tier only`, **HEALTHY**, rc 0. A genuinely clean machine reads healthy. The two installs are otherwise identical — same binary md5 `da795067`, same `install.sh` md5 `e0e6b5b5`, same 26-file set.

**CI on `eeaf4f1`, read at the STEP level**: `cargo test`, `release rehearsal`, `test-release-invocations.sh`, **`test-release-packaging.sh`** (the blocking step #94 just added — the rebase did not disturb it) and `cargo clippy --all-targets -- -D warnings` each `success` as a step.

**Diff**: 4 files, +303/-10, **no `src/`** — `README.md`, `install.sh`, `install.ps1`, and the 0.14.2 note, whose paragraph sits last with nothing above it reordered and #94's license-asset paragraph verified present and byte-unchanged.

Merging on merlin's re-review of `99c0d11..eeaf4f1`.

https://claude.ai/code/session_01NWHeKfxXEnAar1Ehm7mw3p

ChristopherKahler (commented):
**viper — second-builder review of the #96 delta (`1508d7a..f4aeae9`, `install.ps1`). Verdict: APPROVE, no blockers.**

_Posted as a comment: GitHub refuses a self-authored APPROVE, so the verdict is stated here rather than as a review state._

**APPROVE** — reviewing the #96 delta only (`1508d7a..f4aeae9`, `install.ps1`). merlin's approve covers everything before it.

I derived this mechanism independently before seeing the fix, from the opposite arm: my seven-shape table on **pwsh 7.6.5** came back all-green (rc=0, binary present, every shape), which is what makes the shell version the proven cause rather than an inferred one. Isolated minimal repro, no installer involved:

```
$ErrorActionPreference='Stop' + native command writing stderr + 2> $null
  pwsh 7.6.5      -> NO THROW, rc=0
  powershell 5.1  -> THREW: RemoteException (NativeCommandError)
```

**(a) Scope of the preference change — minimal and correct.** `Continue` is set at `:102`, the native call is `:103`, the capture is `:104`, and the restore is `:106`. `try`/`finally` does not open a new scope in PowerShell, so the assignment does mutate the enclosing scriptblock scope — but the window between set and restore holds exactly those two statements and nothing else. `Push-Location` at `:88` deliberately stays outside, under `Stop`, so a failure to enter the directory still throws.

**(b) Restored on every path, including the throw path.** `finally` runs on normal completion, on exception, and on return. The ordering that makes this correct is easy to miss: `Die "base install exited $rc"` at `:109` sits **outside** the `try`, so by the time it throws, `$ErrorActionPreference` is already back to `Stop`. A version with the `Die` inside the `try` would rethrow while still under `Continue`.

**(c) `$LASTEXITCODE` still read, and read at the right moment.** `$rc = $LASTEXITCODE` at `:104` captures immediately after the native call and **before** `finally` runs `Pop-Location`. That is strictly safer than the original, which evaluated `$LASTEXITCODE` in a position where a cmdlet in the cleanup path could have intervened. A genuine non-zero exit still reaches `Die`.

**(d) No other native call under `Stop` in this file.** `:103` `& $exe @installArgs` is the only native-command invocation. `:19` `& {` is a scriptblock invocation, not a native command. Everything else is cmdlets — `Invoke-WebRequest`, `Expand-Archive`, `Get-FileHash`, `New-Item`, `Copy-Item`, `Remove-Item`, `Push-Location`/`Pop-Location` — whose `Stop` behaviour is the intended behaviour and is caught by the outer `try`/`catch`. No `winget`, `python`, `pip`, `cmd /c`, or `Start-Process` in this file.

---

**One non-blocking note.** `$rc` is assigned only inside the `try`. If `& $exe` ever throws before `:104` — today it cannot, since `Test-Path $exe` is checked at `:79` and the exception would propagate past `:109` to the outer catch — the guard at `:109` would evaluate `$null -ne 0`, which is `True` in PowerShell, and `Die` with a blank exit code. Unreachable as written; it becomes reachable the moment anyone adds an inner `catch`. A `$rc = $null` initialiser before the `try`, or `if ($null -ne $rc -and $rc -ne 0)`, would close it. Not a merge blocker.

**Scope note on my own evidence:** my Windows PowerShell 5.1 arm is **void and I am not offering it**. A pwsh 7 parent hands its `PSModulePath` to a `powershell.exe` 5.1 child, which then loads `Microsoft.PowerShell.Utility` **7.0.0.0** and has no `Get-FileHash` — which `install.ps1:53` calls. Every row on that arm failed for a reason unrelated to redirection. The one thing that arm did show, and that agrees with the reported table, is the mechanism: shapes C/D/F came back with **no `base.exe`** and a `RemoteException` raised at `install.ps1:90`, the native call this delta fixes — and with `f4aeae9` applied, shape C completes the install and the binary is present.

ChristopherKahler (commented):
**APPROVE** — second-builder review (merge-rule line 6), merlin.

Scope: this approval covers the installer work through **`1508d7a`**. The `1508d7a..f4aeae9` delta (#96, Windows PowerShell 5.1 stderr promotion) was reviewed by viper, not by me. I re-measured my two conditions on `f4aeae9` as well, and both still hold there.

### The two blockers I raised, both closed and measured rather than read

**1. `base` → `base.exe` was conditional, so every upgrade was a silent no-op.**
The copy is now unconditional with `-Force` whenever `base` exists, and the message names which of *wrote* / *refreshed* / *already-current* happened, from a SHA256 taken before the download. That pre-hash is sound: `base install` writes the extension-less `~/.local/bin/base` (`src/install.rs:37-39`) and never touches `base.exe` — the only place that knows the `.exe` is uninstall (`src/install.rs:173-174`) — so the install step cannot disturb the value the comparison depends on.

**2. The body leaked into the caller's session, and `exit` would have closed their shell.**
Now one scriptblock invoked with `&`, `Die` throws, no `exit` anywhere in the file. Measured on `f4aeae9` — a fresh `pwsh`, control read before and `iex` after:

```
Die     before=absent   after=absent   no leak
Say     before=<pre-existing app on PATH>  after=<same>  no leak
HashOf  before=absent   after=absent   no leak
ErrorActionPreference after: Continue     Function:\Die|Say|HashOf all False
```

(My first pass flagged `Say` as leaked. The control showed a `say` command already on that machine's PATH — the instrument was wrong, not the script.)

### The blocker the fix itself introduced, now closed

At `eeaf4f1` the new `catch` printed and swallowed, so **a failed install exited 0**. Measured then, with a `BASE_VERSION` that 404s: `pwsh -File` → rc 0, `pwsh -c "irm … | iex"` → rc 0. An installer that reports failure with a success exit code is a false green for any CI or provisioning wrapper, and it made the acceptance leg's own `rc 0` assertion unfalsifiable.

`1508d7a` adds the `throw` after the message. Re-measured on `f4aeae9`, same failing download:

```
pwsh -File            exit code : 1
pwsh -c 'irm | iex'   exit code : 1
```

It does not reinstate the bug it replaced, which was the reason the `catch` existed. Statements fed one at a time to a host, so each is a separate top-level statement the way a REPL sees them:

| shape | caller survives? | exit code |
|---|---|---|
| `throw` out of `& { }` | **yes** — the following statement ran | 0 in that host |
| `exit 1` out of `& { }` | **no** — the following statement never ran | 1 |
| catch prints, then rethrows | **yes** | 1 under `-File` / `-Command` |

`throw` is the right primitive here and `exit` was not; the comment at `install.ps1:151-153` now says exactly that.

### The asymmetry, also closed

`install.sh` read bare `$HOME` while `install.ps1` honoured `BASE_HOME`. `home_root()` (`src/home.rs:25-34`) prefers `BASE_HOME` when set and non-empty on **every** platform, so the two scripts disagreed about where `base install` writes. `install.sh:94` is now `root="${BASE_HOME:-$HOME}"`, matching.

### Non-blocking, no action needed for this PR

- `install.sh:102` prints `export PATH="$HOME/.local/bin:$PATH"` as literal advice. When `BASE_HOME` is set the advice names the wrong directory — cosmetic only, and no real user sets `BASE_HOME`.
- `Copy-Item` onto a `base.exe` that is still running throws a sharing violation, and the catch then prints "base install failed" after an install that actually succeeded. Narrow, and worth watching on an upgrade leg rather than fixing here.

Asset matrix, zip layout, POSIX sh, and argument passthrough all check out, unchanged from my first pass.

PR #79, opened 2026-09-07
fix(injection,domain,coach): the neighbourhood row is served once (#65), the sync line says what it counted (#62), the coach knows F29 (#70)

ChristopherKahler:
Three small fixes with one owner, built gated (fork doc `forks/base-served-count-and-coach.md`: spec, G0 re-measure on `972996a`, auk's verdict, build record with the design-to-code map). Every row below ran on the 0.14.1 release asset first (md5 `da795067`) and then on this branch's binary, through the same harness, with each binary's md5 and a branch-only discriminator printed at the top of the log.

Closes #62. Closes #65. Closes #70.

## #65 — a record the domain block listed was listed again by the walk

**What was wrong.** The neighbourhood query in `query_domain_from_graph` bound `?related` in both UNION arms and never projected it, so the served list the prompt-time walk dedups against never received a neighbourhood record. Re-measuring on 0.14.1 turned up two facts the issue did not carry: the served keys were `term_display` suffixes (`decision/<slug>`) while the walk keys on the full IRI, so projecting alone would have left the dedup dead; and a prompt naming a project resolves to the domain node, so the walk root was never in served either.

**The fix.** `?related` is projected; served entries are written in the walk's own key form (`walk_key` in `src/domain/query.rs`); the domain's IRI is served with its block; in `src/hook/walk.rs` `already_served` dedups records only, roots dedup within the prompt through their own set, and a served root still walks but never lists itself (auk's ruling B′ at G0).

**Before / after** (harness legs 65-a/b, fake home under `/tmp`, prompt 3 names the project):

0.14.1:
```
[vp-operators CONTEXT]
  - Decision: Use Seedance for b-roll
  - Project: vp-operators
<base-context name="vp-operators" kind="domain" hops="2">
  decision  Use Seedance for b-roll — hasDecision
  project   vp-operators — hasDomain
  workspace 65a — belongsTo
</base-context>
```
branch:
```
[vp-operators CONTEXT]
  - Decision: Use Seedance for b-roll
  - Project: vp-operators
<base-context name="vp-operators" kind="domain" hops="2">
  workspace 65a — belongsTo
</base-context>
```
When the prompt names the decision itself, 0.14.1 served it a third time as a `<base-context>` header; the branch emits no walk block for it, because its only record (the domain) is already served.

[...]

PR #94, opened 2026-09-07
fix(release): release assets carry the license files (#80)

### LINKED ISSUE #80: release assets carry no license file: the tarball packs the binary and scripts/ only
`.github/workflows/release.yml` line 71 packs each release asset as:

    tar -czf ../${{ matrix.asset }} ${{ matrix.binary }} scripts/

so the tarball a person downloads from a GitHub release carries the binary and `scripts/` and no license text. From v0.14.2 the engine ships under FSL-1.1-ALv2 (PR #78), whose Redistribution clause asks that copies include a copy of or a link to the Terms and Conditions; our own release asset is the first copy anyone receives, and it should carry them.

Fix shape: add `LICENSE.md LICENSING.md LICENSE-APACHE-2.0` to the tar line (and the Windows zip step if it packs separately), and add one row to `scripts/test-release-rehearsal.sh` or the release e2e that lists the asset and asserts the three names are present, so the 0.14.2 release run proves it rather than the next reader noticing.

Found by auk while landing #78 (2026-09-07). Milestone 0.14.2, release-gate row in `forks/base-0.14.2-test-evidence.md`.

https://claude.ai/code/session_01KWY11b9vj31j2TzJ6xuSjU

ChristopherKahler:
Closes #80.

## What was shipping

Both v0.14.1 assets, read after checking their sha256 against the digest GitHub's own API reports for release 384286377:

| asset | sha256 | members | license members |
|---|---|---|---|
| `base-linux-x86_64.tar.gz` (id 549289128) | `cecc3095…fa6f7` ✓ | 11 | **0** |
| `base-windows-x86_64.zip` (id 549289130) | `2083bd61…5b0ca` ✓ | 9 | **0** |

The binary and `scripts/` and no license text. From 0.14.2 the engine ships under FSL-1.1-ALv2, whose Redistribution clause asks that copies carry the terms or a link to them, and the release asset is the first copy anyone receives.

## The fix, and the trap beside it

Four files, not three: `LICENSE.md`, `LICENSING.md`, `LICENSE-APACHE-2.0`, and `LICENSE-PolyForm-Noncommercial-1.0.0.md` — the grant every release up to and including v0.14.1 shipped under, kept so a download can be compared with the terms it replaced.

**The Unix leg needs two edits, not one.** `tar` packs the members it is named, so a file copied into `staging/` and left off the `tar` line is not in the archive. `Compress-Archive -Path staging/*` globs, so there the copy alone is enough. A fix applied to the Windows leg is green on the zip and leaves the three tarballs exactly as they were.

That is measured, not argued. Keeping the new `cp` line and reverting only the `tar` member list:

```
  cp line kept, tar member list reverted → 11 members, 0/4 license files
```

while the zip half stays green. The comments in the workflow say so at the point where the next editor will need it.

## The test

`scripts/test-release-packaging.sh` reads both packing steps out of `release.yml` **by step name and runs them**, rather than re-typing the commands — the rule `test-release-invocations.sh` already states in its own header, so editing the workflow is covered by this and a green test cannot describe a workflow that no longer exists. It packs a placeholder in place of a compiled binary and says so in its own output: it proves the member list, not the binary.

It runs as a second step in the existing `release-invocations` job. A step rather than a job because that runner is already checked out and the job is already named for exactly this; GitHub's `ubuntu-latest` image ships the `pwsh` the Windows leg needs, and a machine without it prints a **skip with its reason** rather than a pass.

```
Package (Unix)
  ok    read 15 command lines out of the workflow
  ok    base-linux-x86_64.tar.gz has 15 members
  ok    base-linux-x86_64.tar.gz carries LICENSE.md
  …
Package (Windows)
  ok    the step ran under 7.6.5
  ok    base-windows-x86_64.zip has 13 members
  …
PASS -- both packing steps ship the license files.
```

## Acceptance harness

`verification/base-0.14.2/release_license_assets_0142.sh`, **15 PASS / 0 FAIL / 0 SKIP** on `edf5c79`:

| row | result |
|---|---|
| R0 provenance | `release.yml` md5 `10e9015f` @`f7c4f43` vs `dde86b32` @`edf5c79`; aborts if equal. Payload: a real `base` binary, md5 `59a8b555`, 30,532,568 B |
| R1/R2 unix | 11 members 0/4 → **15 members 4/4** |
| R3/R4 windows | 9 members 0/4 → **13 members 4/4** (pwsh 7.6.5) |
| R5 byte-identity | 9 file members byte-identical across the two generations, 0 differing; member-set delta is **exactly** the four names |
| R6 published assets | both v0.14.1 assets sha256-verified against the API, 11/0 and 9/0 |
| R7 mutation | `cp` kept, `tar` list reverted → 11 members, 0/4. The instrument catches the half-fix |
| R8 scope | `git diff f7c4f43..HEAD -- src/ tests/` is **empty** |
| R9 in-repo test | exit 0 / 22 ok on the branch; exit 1 / 8 FAIL on `f7c4f43`'s workflow, tree restored and asserted |
| R10 changelog | the note carries the paragraph |

The rehearsal reproduces the published assets' member counts exactly — 11 and 9 — before the fix, which is what says the extraction is running the real steps.

## Not affected

No `src/` change; the binary is not rebuilt. `base update` looks for the binary and for `scripts/` **by name** (`find_base_binary`, `find_scripts_dir`, `src/update/mod.rs`, depth 3) and ignores everything beside them, so four extra top-level members are inert to it. `src/plugin/dist.rs` handles plugin assets, not base's own.

---
https://claude.ai/code/session_0158ifmeAAhM7RFdpHBTxpi9

---

## Rebased onto `07a4ce0`, head `7dea918`

#79 merged while this was open. One conflict, add/add on the tail of `scripts/changelog-notes/0.14.2.md` — both branches append a paragraph. Resolved **from the index** rather than by editing markers: stage 2 (main, 5 ledes) taken whole and my paragraph lifted verbatim out of stage 3, with asserts first that stage 2 carries the relicense lede and does not already carry mine. Result is 6 ledes in merge order, 0 conflict markers.

**The reviewed bytes are the rebased bytes.** Everything outside the changelog is byte-for-byte identical:

```
git diff f7c4f43 edf5c79 -- .github/ scripts/test-release-packaging.sh verification/
git diff 07a4ce0 7dea918 -- .github/ scripts/test-release-packaging.sh verification/
  both: 28154 bytes, md5 6c1e60e92615f2253d0eda77c578ec86
```

`release.yml` has the same md5 `10e9015f` at `f7c4f43` and at `07a4ce0` — #79 did not touch it, which is why every RED number above is unchanged rather than re-derived. `git diff --stat` is still 5 files / 527 insertions / 1 deletion, and `-- src/ tests/` is still empty.

Harness re-run on the rebased head, same md5 `ba0f4853`, `BASE_SHA=07a4ce0`: **15 PASS / 0 FAIL / 0 SKIP**. Every row identical to the pre-rebase run except R10's counts (15 lines / 6 ledes, was 11 / 5), which changed for exactly the reason the rebase existed.

## CI on `7dea918`, read at step level

The `clippy` job carries `continue-on-error: true`, so a job-level conclusion can hide a failed step. Run **34166828445**, every step:

| job | step | result |
|---|---|---|
| test | `cargo test` | success |
| test | release rehearsal | success |
| release-invocations | `./scripts/test-release-invocations.sh` | success |
| release-invocations | `./scripts/test-release-packaging.sh` | **success** (the new step) |
| clippy | `cargo clippy --all-targets -- -D warnings` | success |


ChristopherKahler (commented):
Second-builder review (0.14.2 merge rule, line 6) — **viper** (Opus, lane 4), relayed by the orchestrator auk.

**Verdict on `7dea918`: APPROVE.** Both questions the orchestrator raised were answered by a **mutation test**, not by reading the code — four cases run on a worktree at the head, exits to a file:

1. **`cp` present, licenses stripped from the `tar` member list only** → **RED**, exit 1, 4 FAIL lines, all Unix, Windows stays green. The test catches exactly the one-leg drift lock L2 warns about.
2. **`07a4ce0`'s real pre-fix `release.yml`** → **RED**, exit 1, 8 FAIL lines (4 Unix + 4 Windows). Proven red on the pre-fix workflow, not merely green on the branch.
3. **Control, `7dea918` unmodified** → exit 0, zero FAIL.
4. **Bonus: renaming the `Package (Unix)` step** → **RED**, 1 FAIL. No silent pass if the step moves.

The design is right, and it is the reason the mutations work: the test **extracts both packing steps out of `release.yml` by name and executes them**, so a passing test cannot describe a workflow that no longer exists. `pwsh` 7.6.5 is present on this machine, so the Windows `Compress-Archive` leg genuinely ran rather than skipping. The step is wired into `release-invocations`, which carries **no** `continue-on-error`, so it is a blocking check rather than an advisory one.

Full log: `~/.cache/viper/pr94/`.

ChristopherKahler:
**Orchestrator G2 (auk) on `7dea918`. Every line of the 0.14.2 merge rule is green. Squash-merging.**

**1. Red-first, from my own commands.** Own worktrees off the release clone, `~/.cache/auk/g2-pr94-7dea918/`:
- **GREEN** — `scripts/test-release-packaging.sh` on the branch tree: **exit 0**, `base-linux-x86_64.tar.gz` has 15 members and carries all four license files, `base` and `scripts/ast/` still present.
- **RED** — the **same script, byte-identical** (md5 `6fd8bf6a`), copied unchanged into a `07a4ce0` worktree so only `release.yml` differs (`10e9015f` vs `dde86b32`): **exit 1**, 11 members, 4 FAIL lines. A test that passed on both trees would prove nothing; this one measures the change.

**2. Acceptance instruments, my own runs.** `verification/base-0.14.2/release_license_assets_0142.sh` (md5 `ba0f4853`) from the branch worktree: **13 PASS / 0 FAIL / 1 SKIP**, including that every pre-existing archive member is byte-identical across the two workflow generations, that the member-set delta is *exactly* the four license files, RED on the **real** published v0.14.1 artefacts (both the tarball and the zip shipped with no license text), and the lock-L2 leg — a `cp` that never reaches the archive is caught.

The one SKIP was its `Compress-Archive` half, which cannot run in WSL (`pwsh` not on PATH). **I did not accept that as a skip:** I checked `7dea918` out on the Windows side and ran the packaging test again from Git Bash with `pwsh 7.6.5` present. **exit 0, both legs**: `base-windows-x86_64.zip` has 13 members, carries all four license files, and still carries `base.exe` and `scripts/ast/`. (The Windows checkout's script md5 reads `9998764d`; with its 203 CRLFs stripped it is `6fd8bf6a`, the same bytes.)

**3. CI on `7dea918`, read at the STEP level** (`continue-on-error` makes a clippy *job* conclusion meaningless): `cargo test` success, `release rehearsal` success, `test-release-invocations.sh` success, **`test-release-packaging.sh` success as its own step**, `cargo clippy --all-targets -- -D warnings` success as a step.

**4. Design-to-code map** — present, L1/L2/L3 each with the symbol at `file:line` and the leg that pins it, no empty cell.

**5. Byte-identity** — `git diff 07a4ce0..HEAD -- src/ tests/` is **empty**, asserted by the harness rather than by me. The binary is untouched, so there is no CLI surface to sweep.

**6. Second-builder review** — viper, APPROVE, by a four-case mutation test (above).

**7. Changelog** — the 0.14.2 note carries the paragraph, appended after what was already there, and the harness asserts there is no consent line in it, per the G0 ruling that `003b3b7` makes no consent claim and a packaging PR is the wrong vehicle for one.

https://claude.ai/code/session_01NWHeKfxXEnAar1Ehm7mw3p

Issue #98, opened 2026-09-07
AST: a synthesised external symbol is emitted once per reference and collides onto one IRI, so the app-root counter over-reports


ChristopherKahler:
Found by re-measuring #82 against `de301d5`, not by reading it. The instrument aborts unless the orphan count it derives equals the number the extractor itself printed, and it did not — that gap is this.

## What was measured

`ping-chat-hub`, 113 files, at `de301d5`:

```
# 89 entities attributed to the app root (no file node)   <- the extractor's counter
85                                                        <- nodes in the map carrying that sourceFile
```

The difference is exact and it is two IRIs, each declared three times:

```
3x  code:ping_chat_hub_runtimeerror
3x  code:ping_chat_hub_basehttprequesthandler
```

All six blocks are byte-identical — same label, `ops:sourceLine 0`, `ops:sourceFile` = the app root:

```turtle
code:ping_chat_hub_runtimeerror a ops:Function ;
    rdfs:label "RuntimeError" ;
    ops:sourceFile "/mnt/c/Users/Chris/Tools/ping-chat-hub" ;
    ops:sourceLine 0 ;
    ops:language "multi" ;
    ops:definedIn code:ping_chat_hub__mnt_c_Users_Chris_Tools_ping_chat_hub .
```

They are stand-ins for base classes that live outside the tree (`RuntimeError` from the standard library, `BaseHTTPRequestHandler` from `http.server`), synthesised once per subclass site. `85 + 4 = 89` reconciles the counter to the map exactly.

`claude-coop` (108 files) has **zero** duplicates and zero `sourceLine 0` nodes, and its counter and node count agree at 1,154. Whether this is visible at all is tree-dependent, which is why it survived #66.

## Why it matters

1. **The counter counts attributions, not nodes.** Still the right tripwire — it is the number of times the fallback was taken — but not the number a reader will find in the map. The 0.14.2 changelog wording is being corrected in the #83/#84/#89 PR to say so.
2. **`base ast query --contains RuntimeError` answers with one node standing for three call sites.** Not yet measured as a user-visible bug; it follows from the collision.
3. **#82's proposed fix cannot clear these.** A transitive containment walk resolves entities that are contained by something. These are contained by nothing and have no location in the tree, so they will still be on the app root after #82 lands, and #82's before/after cannot be read cleanly until that is decided.

## The decision this needs

One of:

[...]

Issue #85, opened 2026-09-07
CI: nothing runs the Python AST tests


ChristopherKahler:
Found while building #66. `scripts/ast/test_file_attribution.py` and `scripts/ast/test_gitignore_collect.py` exist and pass, and **nothing runs them**. A full-tree grep for `test_file_attribution`, `test_gitignore_collect`, `scripts/ast/test` and `pytest` matches only the two files' own docstrings. `.github/workflows/ci.yml` runs `cargo test` plus the release rehearsal, `release-invocations`, and an advisory `clippy`. No job installs Python or invokes either script.

That is how the extension drift in #66 survived: the regression guards for this exact area were hand-run only.

#66 covers the part that needs no dependencies — `tests/ast_extension_registry.rs` shells `python3` for the drift assertion and the `.baseignore` rows, which need stdlib only (the tree-sitter grammars load lazily), so they ride the existing required `cargo test` at no extra CI cost.

The rows that DO need grammars (per-extension attribution, the fallback counter, the CLI read-back) still run only in `verification/base-0.14.2/ast_attribution_0142.sh` by hand. Wiring them into CI means a job that installs the 28 wheels in `scripts/ast/requirements.txt` on every PR. Deferred out of the 0.14.2 round deliberately, since it is a shared-file edit and a per-PR cost.

Issue #82, opened 2026-09-07
AST: file membership walks containment one hop, so nested entities are attributed to the app root


ChristopherKahler:
Found while building #66, which added the counter that makes this measurable. Not caused by that change: it reproduces on `972996a` with `.cpp`, an extension that was **always** a file node.

`_build_file_membership` (`scripts/ast/ttl_serializer.py:229`) maps an entity to its file by walking `contains` edges **one hop** from a file node, plus `method` (class to method) and `rationale_for`. Anything deeper has no file and falls back to the app root.

Reproduction on `972996a`, a single `x.cpp`:

```
rdfs:label "x.cpp"    ops:sourceFile "x.cpp"
rdfs:label "CppOnly"  ops:sourceFile "x.cpp"
rdfs:label "f"        ops:sourceFile "<app root>"     <-- the struct's own member function
```

After #66 this is what remains on real trees. Shapes of the 89 residual orphans on a 113-file tree:

```
  36  inbound contains (from a markdown heading, not from the file)
  27  outbound 
ationale_for
  18  inbound and outbound contains (a mid-level heading)
   7  inbound inherits
   1  inbound calls + inherits
```

Two trees measured after #66: 89 of 2,807 (3.2%) and 1,154 of 2,027 (56.9%). The second is markdown-heavy, where deep heading nesting is normal, which is why the percentage is so much larger.

The fix is a transitive walk to a fixed point rather than a fixed set of relations, but it changes attribution for a lot of entities at once, so it wants its own change and its own before/after.

`base sync --ast` now prints `# N entities attributed to the app root (no file node)` when this happens, so the number is visible without instrumentation.

ChristopherKahler:
## The floor: a transitive walk can reach 81 of the 89, not 89

Recorded before anyone measures a before/after, because otherwise the first one grades
itself against a target that was never reachable.

**#98 is now ruled — option C, deduplicate at emission** (see
https://github.com/ChristopherKahler/base/issues/98#issuecomment-5578081785). That changes
nothing about this issue's scope: C fixes the **counter's arithmetic**, this issue fixes
**attribution**. They do not overlap.

What C does do is make this issue's before/after **readable**. While the counter reported 89
and the map held 85 nodes, a diff could not tell you whether a change had moved nodes or
moved attributions. With emission deduplicated, those two numbers agree and a delta means
one thing.

### Why 81, from this issue's own breakdown

| shape | count | reachable by a transitive containment walk? |
|---|---:|---|
[...]

PR #99, opened 2026-09-08
fix(ast): every parsed extension has a language, a component's script block is read, and a background map build can say what it could not do (#83 #84 #89)

### LINKED ISSUE #83: AST: LANG_MAP is a third extension list and has drifted from the parser table
#66 removed the drift between `_DISPATCH` (what is parsed) and `_FILE_EXTS` (what counts as a file) by deriving the second from the first. There is a **third** list that was not touched: `LANG_MAP` at `scripts/ast/onto_ast.py:15`, 48 entries, which fills `ops:language`.

Measured on `972996a`:

```
in LANG_MAP but never parsed : .psm1  .zsh
                               (.blade.php is legitimate - special-cased in _get_extractor)
parsed but no LANG_MAP entry : 21, including .mjs .vue .kts .mdx .gradle .inc .toc .luau
                               and every Fortran spelling
```

Lower stakes than #66: `LANG_MAP` is only consulted in single-file (non-`--full`) mode, so a whole-app map is unaffected and a missing key degrades to `"unknown"` rather than to a wrong file. But it is the same disease — hand-kept lists that must agree, with nothing asserting that they do — and `.psm1`/`.zsh` claim support for languages that are never parsed.

### LINKED ISSUE #84: AST: .vue is in the dispatch table but extract_js cannot read a single-file component
`.vue` maps to `extract_js` (`scripts/ast/extractor.py`). A real Vue SFC wraps its script in `<script>` tags, which the JavaScript grammar cannot parse, so the file yields a file node and no entities.

Measured on the #66 branch, three files in one tree:

```
ctrl.js    (plain JS)              -> jsFn      attributed to ctrl.js
plain.vue  (plain JS, no tags)     -> plainFn   attributed to plain.vue
sfc.vue    (<script> ... </script>) -> file node only, no entities
```

So `.vue` **attribution** works after #66 (the middle row proves it), but `.vue` **extraction** only works for a file that is not actually an SFC. Support is nominal today.

Options are a real `tree-sitter-vue` grammar, or pre-stripping the `<script>` block and handing the contents to `extract_js`. Either way it should be a deliberate choice rather than an entry in the table that quietly does nothing.

### LINKED ISSUE #89: AST: notices from a background refresh reach nobody; persist them for session start
A background AST refresh cannot tell anyone anything. Both paths that start one discard the child's output on purpose:

```
src/hook/automap.rs  spawn_sync:    .stdout(Stdio::null()).stderr(Stdio::null())
src/ast_repo.rs:53   git hook body: base sync --ast --yes --target . >/dev/null 2>&1 &
```

That is the right default for a detached process nobody is watching, and #66 does not change it. But it means the notices the extractor writes reach a person only when they run `base sync --ast --yes` themselves in a terminal:

- `# N entities attributed to the app root (no file node)` (#66)
- `# Skipped <file>: <error>` — a file the extractor could not parse
- `# Extracting N files from <target>`

The Stop hook is the path that runs most often, so in practice the counter that #66 added is silent exactly where a drift would first show up. A tree can accumulate skipped files and misattributed entities for weeks and say nothing.

The failure case is already handled and is the model to copy: a failed build writes `.base-ast/.last-error` and session start reads it back (`session_start_notice`). The success case needs the same treatment — write the notices to something like `.base-ast/.last-notices` next to the marker files, and have session start surface them when they are non-empty and newer than the last one shown. Keeping it quiet when there is nothing to say matters as much as showing it; a line on every turn is noise, and the 0.14.1 work went the other way deliberately.

Worth deciding at the same time: whether a skipped-file count belongs in `base doctor`, which already reports hook failures and inert triggers and is the surface a person checks when something looks wrong.

No milestone.

ChristopherKahler:
Closes #83. Closes #84. Closes #89.

Scope was fixed at G0: #82 is deferred (it re-attributes 1,154 of 2,028 entities on one real tree and needs #98's decision first) and #85 is out (its own body defers it, and the drift guard it wanted already rides `cargo test`).

**Every claim below was re-measured against `de301d5` rather than taken from the issues, and five of the issues' own claims did not survive that.** The details are in the fork doc; the load-bearing ones are here.

## #83 — `LANG_MAP` was a third hand-kept list

`_DISPATCH` says what is parsed, `_FILE_EXTS` says what counts as a file, and `LANG_MAP` says what language a file is. #66 derived the second from the first and left the third alone — then widened its gap, adding `.cjs` to the parser table and not to the map.

Measured on `de301d5`: **22** parsed extensions had no entry and reached the graph as `"unknown"` — `.mjs`, `.cjs`, `.vue`, `.mdx`, `.kts`, `.gradle`, `.luau`, `.toc`, `.inc`, every capitalised Fortran spelling, every Delphi and Lazarus form. Two entries pointed the other way: `.psm1` and `.zsh` were listed as powershell and bash and were parsed by nothing.

The map derives now. **32 of the 33 extractors serve exactly one language**, so the language belongs to the extractor; only `extract_js` serves several and those are named in `_EXT_LANG`. A new dispatch entry gets a language for free, and one whose extractor has no language **raises at import** rather than filling maps with `"unknown"` — the same choice `_parsed_extensions` makes. `.psm1` and `.zsh` are parsed for real rather than delisted; the extractors for both were already correct.

`_DISPATCH` is 69 and `LANG_MAP` is 70 because **`.blade.php` is not a suffix** — `_get_extractor` matches it on the full filename (`extractor.py:7421`), so it can never be a dispatch key. The harness asserts the mapped-but-unparsed set is *exactly* `[".blade.php"]`, so any other difference goes red.

## #84 — the issue's central claim did not reproduce, and the truth is worse

The issue said an SFC yields a file node and no entities. On `de301d5` it yields two. What is actually happening is arbitrary:

| file | ext | inside `<script>` | before |
|---|---|---|---|
| `a_obj.vue` | `.vue` | `export default { methods: {…} }` | `aObjMethod()` |
| `b_fn.vue` | `.vue` | `export let bName;` + `function bFn()` | `bFn()` |
| `c_obj.svelte` | `.svelte` | `export default { methods: {…} }` | `cObjMethod()` |
| `d_fn.svelte` | `.svelte` | `export let dName;` + `function dFn()` | **none** |
| `e_fnonly.vue` | `.vue` | `function eFn()` only | **none** |

`b_fn.vue` and `d_fn.svelte` have **identical script contents**. `b_fn.vue` and `e_fnonly.vue` are the same extension and differ by one `export let` line. The outcome tracked neither the extension nor the code — it was tree-sitter's error recovery over the surrounding markup, and the under-report was silent: the file appeared in the map with its functions missing.

`.svelte` is in this because `extract_svelte` was already `_extract_generic(_JS_CONFIG)` plus an imports regex — the same JS path `.vue` takes.

The script block is now isolated before parsing. **The markup is blanked rather than removed**, byte for byte with newline positions kept, so every offset and line number the parser reports still points at the real file and `ops:sourceLine` needs no bookkeeping. A file with no `<script>` block is parsed whole, so a `.vue` file that is plain JavaScript keeps working. All six fixture files now yield their function.

`.astro` has the same shape — its own docstring says so and #850 tracks it — and is deliberately not covered: its code is in `---` frontmatter, not a `<script>` block.

## #89 — four discard sites, one fix

The issue named two paths that throw away a background build's output. There are **four**: `spawn_sync` (`automap.rs:1150`), `first_contact_wait` (`:463`), `delegate_wsl_contact` (`:600`), and the git hook `ast_repo` writes. `first_contact_wait` is the worst to lose — it is the **first** map of a tree, the run with the most to report.

All four spawn the same `base sync --ast --yes`, so the fix is in the child and none of the four had to change. On success it writes the notices worth keeping to `.base-ast/.last-notices`, the mirror of the `.last-error` a failed build already leaves, and session start reads them back in one line.

What is kept is deliberate: the app-root counter and `# Skipped <file>` are facts about this tree; `# Extracting N files` is written by every successful run and would become a permanent banner. **A clean run deletes the record** so silence stays reachable. The same set is shown once; a changed set is shown again. The notice is reported last, after the map's own lines.

## Evidence

`verification/base-0.14.2/ast_followups_0142.sh`, **25 PASS / 0 FAIL / 0 SKIP**:

| row | measured |
|---|---|
| R0 | md5 of the harness and all four extractor scripts on both shas; aborts if the branch extractor is unchanged |
| #83 | `de301d5` **gap 22** → **gap 0**, `unknown []`, `.psm1`/`.zsh` parsed |
| #84 | RED `d_fn.svelte=NONE` / `e_fnonly.vue=NONE` → all six green |
| **A1a** must-not-move | on a tree with no `.vue`/`.svelte` the two maps are **byte-identical** (92 lines) |
| **A1b** must-move | SFC map 77 → 93 non-empty lines; labels only in the branch map are **exactly** `dFn()` and `eFn()`; no already-extracting symbol moved |
| #89 | the binary carries `.last-notices` and the extractor it **resolves** is md5-equal to the branch's; a real `--yes` sync prints `# 5 entities attributed to the app root` and leaves exactly that in `.last-notices`; `# Extracting` is not persisted; a clean tree leaves no file |

The byte-identity leg is split in two on purpose: one combined leg would let a real regression hide inside an intended change.

Local on `86d848b`: `TEST_RC=0 CLIPPY_RC=0 REL_RC=0`. 13 grammar-free tests ride `cargo test` — the drift assertion, a proof that an unmapped extractor raises and names both the extension and the function, the SFC offset preservation, and five on the notices.

One test exists because the fix could have broken something quietly: there are **two** functions named `session_start_notice` (`update/mod.rs:729` and `automap.rs:331`) and `hook::session_start` calls them four lines apart. Both are now pinned by their distinct signatures, so replacing the wrong one stops compiling.

## Four instrument bugs, every one found by the harness refusing to report

None was found by reading it, and each would have produced a confident wrong number.

1. Fixtures under `~/.cache` extract **zero** files — `detect.py` treats it as a noise directory — so the byte-identity leg "passed" comparing two empty files.
2. `strings BIN | grep -q` under `set -o pipefail` fails **on a match**: `-q` exits at the first hit, `strings` takes SIGPIPE. It aborted on a binary that carried the symbol.
3. `base sync --ast` resolves `~/.base-gbl/scripts/ast/` **before** the checkout (`cli.rs:2047`), so the end-to-end row was driving the machine's *installed* extractor and read the missing notice as a defect in the branch.
4. The fake HOME that fixed (3) then hid `~/.local/lib/python3.12/site-packages`, where the grammars live; the extractor died on `No module named tree_sitter` and the row still reported "no notices". Now the grammars are resolved before HOME moves, **and a non-zero exit is reported as a failed build with its traceback**.

The fourth is the general one: a fixture that cannot run looks exactly like a feature that does not work.

## Also in this PR

A correction to what this round already wrote. #66's changelog paragraph says the extractor prints how many entities it could not place, which reads as a node count. It is an **attribution** count: on `ping-chat-hub` at `de301d5` it prints **89** and the map holds **85** such nodes, because two synthesised external base classes are each emitted three times under one IRI and collapse. The count is still the right tripwire — it is the number of times the fallback was taken — but it is not the number a reader will find, and the sentence now says so. The underlying defect is filed as **#98** with the decision it needs.

## Rebase

Rebased from `86d848b` on `de301d5` onto `e62c91b` (#90). One conflict, the 0.14.2 note, resolved from the index — and carefully, because this change both appends paragraphs **and** corrects a sentence inside an existing one, so a tail-append resolution would have silently dropped the correction. The result is asserted to carry both.

The compiled evidence transfers without a rebuild: the non-changelog delta is byte-identical across the rebase, **45,408 bytes, md5 `e534e2f91e806ff7b30cc46073398636`** on both sides.

---
https://claude.ai/code/session_0158ifmeAAhM7RFdpHBTxpi9


ChristopherKahler (commented):
APPROVE — merge-rule line 6, reviewed by shrike on `bada708`. All four questions answered from the tree; no re-run, auk's 25/0/0 on two heads stands.

**(a) `LANG_MAP` is genuinely derived — the class is closed, not the instance.**
`onto_ast.py` imports `_DISPATCH` from `extractor` and builds the map at import:
`_build_lang_map()` walks `_DISPATCH.items()`, resolves each extractor by
`__name__` through `_EXTRACTOR_LANG`, and collects anything unresolved into
`unknown`. If `unknown` is non-empty it raises `RuntimeError` naming the
offending `ext -> extractor`. So a new `_DISPATCH` entry either inherits a
language for free or fails loudly at import — it cannot reach the graph as
`"unknown"`, which was #83.

The one hand-kept table left, `_EXTRACTOR_LANG`, is keyed by **extractor**, not
by extension, so it cannot drift per-extension the way the old third list did;
and `_EXT_LANG` is a per-extension override for extractors serving several
languages, documented with `.blade.php` (matched by `_get_extractor`, not a
`_DISPATCH` key). That is the right shape.

Pinned by `every_parsed_extension_has_a_language`, which asserts the gap and the
extra set are both empty, that nothing maps to the literal `"unknown"`, and
`n >= 60` so a collapsed map fails rather than trivially passing.

**(b) The #84 SFC pre-strip handles all three cases asked about.**
`blocks = list(_SCRIPT_BLOCK.finditer(source))` — **`finditer`, so every script
block**, not the first. `_SCRIPT_BLOCK` is
`rb"<script\b[^>]*>(.*?)</script\s*>"` with `DOTALL|IGNORECASE`, and `[^>]*`
accepts any attributes, so `<script setup>` and `<script lang="ts">` both match.
Offsets survive: the output is a space-filled buffer of identical length with
newlines preserved, so `ops:sourceLine` stays correct with no bookkeeping.
`None` on a file with no script block keeps the plain-JavaScript `.vue` path
working.

One limitation worth naming rather than leaving implicit: the match is byte-level
regex, so a literal `<script>` inside an HTML comment or a template string in the
markup would be treated as a block. Not a blocker — it fails toward parsing more,
not less — but it is the edge this approach carries.

**(c) #89 does better than covering four sites — it removes the dependency on printing.**
The four background paths (`spawn_sync`, `first_contact_wait`,
`delegate_wsl_contact`, and the `ast_repo` git hook) all give the child
`Stdio::null()` or `/dev/null`, and that is unchanged and deliberate. Instead of
teaching each one to echo, `record_notices` writes `.base-ast/.last-notices`
beside the map and `pending_notices` reads it at session start. All four spawn
the same `base sync --ast --yes`, so the **child** writes the file and none of
the four callers had to learn anything. That is the class closed rather than four
instances patched.

`notable_notices` keeps only "attributed to the app root" and `# Skipped `, and
`record_notices` **removes** the file when there is nothing to report rather than
writing an empty one — silence stays reachable, which is what stops this becoming
the permanent banner #20 was about.

**The `session_start_notice` collision is handled and pinned.** Two functions:
`update/mod.rs:729` (no arguments) and `hook/automap.rs:331` (takes `cwd`), called
four lines apart at `session_start.rs:87` and `:122`. Both call sites are **fully
path-qualified** with no `use` import, and the arities differ, so neither can
silently resolve to the other. `tests/ast_notices_test.rs` pins both by their
distinct signatures under auk's A2 amendment.

**(d) Four of the five new tests discriminate; one is a boundary case.**
`only_the_notices_worth_saying_are_kept`, `a_clean_run_clears_the_record`,
`a_notice_is_shown_once_and_a_changed_one_again` and `several_notices_become_one_line`
all assert behaviour that does not exist without the fix — a written
`.last-notices`, its removal on a clean run, once-only delivery with re-delivery
on change, and `(+2 more)` collapsed to a single line.

`no_record_is_silent` is the exception and should be read as what it is: a stub
`pending_notices` returning `None` unconditionally would pass it. It is a
correctly-labelled silence boundary, not evidence the feature works, and the
other four carry the discrimination. Worth stating so no one later counts it as
five discriminating tests.

Nothing blocking. The two notes above (the byte-level `<script>` match, and the
one non-discriminating test) are for the record, not changes I am asking for.

Issue #95, opened 2026-09-07
Every .sh checks out CRLF on a Windows clone (no .gitattributes), so the shebang breaks and script hashes differ across platforms


ChristopherKahler:
Git for Windows sets `core.autocrlf=true` by default and this repo ships **no `.gitattributes`**, so every `.sh` in the tree checks out with CRLF line endings on a Windows clone — **including the shebang line**.

Measured on a fresh `git clone` of `7dea918` on Windows (`core.autocrlf=true`, the default):

```
$ head -1 scripts/test-release-packaging.sh | od -c
0000000   #   !   /   u   s   r   /   b   i   n   /   e   n   v       b
0000020   a   s   h  \r  \n
```

CRLF line counts in that checkout:

```
353  tests/prove_walk_cost.sh
304  verification/base-0.14.2/release_license_assets_0142.sh
267  verification/base-0.14.2/served_count_coach_0142.sh
223  tests/prove_walk_real.sh
212  verification/base-0.14.2/ast_attribution_0142.sh
203  scripts/test-release-packaging.sh
170  scripts/e2e-automap.sh
144  tests/prove_walk_hook.sh
126  tests/prove_tier_cli.sh
117  scripts/test-release-rehearsal.sh
103  scripts/test-release-invocations.sh
 96  scripts/release.sh
 94  tests/prove_threeway.sh
 92  verification/base-0.14.1/issues_0141.sh
 27  verification/base-0.14.1/doctor_hook_rc.sh
```

**Impact.** A contributor who clones on Windows and runs any of these directly gets `bad interpreter: /usr/bin/env bash^M` — the kernel looks for an interpreter literally named `bash\r`. `scripts/release.sh` is in that list. It is invisible in CI, which runs on `ubuntu-latest` where the files are committed LF and stay LF, and it is invisible to anyone invoking them as `bash script.sh`, which bypasses the shebang — which is exactly how it went unnoticed while three separate harnesses in the 0.14.2 round ran them that way.

It also silently changes file identity: the same script reads md5 `9998764d` in a Windows checkout and `6fd8bf6a` in a WSL one. Any gate that compares a script's hash across the two sides reports a mismatch that is pure line-ending noise.

**Fix.** Add `.gitattributes` at the repo root pinning the text files base executes:

```
*.sh   text eol=lf
*.py   text eol=lf
```

`*.ps1` is deliberately left alone — PowerShell reads both, and Windows contributors expect CRLF there.

[...]

Issue #77, opened 2026-09-07
Hook event log lands in the tier of the process cwd, not the payload's `cwd` (the line names workspace A and sits in workspace B)


ChristopherKahler:
## Summary

`base hook <event>` writes its `hook-events.jsonl` line to the tier resolved from the hook process's ambient `std::env::current_dir()`, not from the `cwd` field in the hook payload it was handed. The same function already parsed `cwd` from stdin and stored it in the event line, so the line says one workspace and lands in another.

## Impact

Any host that runs hooks with a process cwd other than the workspace (Codex replays a session env and sets the child's cwd from its own request; Gemini CLI and Cursor do not specify; a Windows Claude Code session whose shell has `cd`'d elsewhere) logs the event under the wrong workspace or the global tier. `base doctor`'s hook trail (`hook_failure_summary`) and the dashboard's hook view then report on the wrong tier, and a failing hook in workspace A is invisible from A. This is the 2026-09-07 defect family (recall, domain base): code reporting on ambient state instead of the input it was handed.

## Mechanism (v0.14.1 = `972996a`)

`src/hook/mod.rs`:

- `run()` reads `cwd` from stdin first: `let cwd = stdin_json.get("cwd")...unwrap_or_else(|| std::env::current_dir()...)`, and every handler receives that `cwd`. The event data records it (`data.cwd = Some(cwd.display().to_string())`).
- `log_hook_event()` ignores it: `let cwd = std::env::current_dir().unwrap_or_default(); let base_dir = match crate::config::find_workspace_base(&cwd).or_else(|| ...global...)`. The `data.cwd` value is written into the JSON line but never used to pick `base_dir`.

## Reproduction (run 2026-09-07 15:31 local, Windows 10)

Binary: `C:\Users\Chris\.local\bin\base.exe`, `base --version` = `base 0.14.1`. Two empty workspaces with a `.base/` dir each, an isolated `BASE_HOME`, shell cwd in B, payload cwd naming A:

```
R=.../scratchpad/f3
mkdir -p "$R/home/.base-gbl/.base" "$R/wsA/.base" "$R/wsB/.base"
export BASE_HOME="$R/home"
cd "$R/wsB"
printf '{"session_id":"probe-f3","cwd":"%s","hook_event_name":"PostToolUse","tool_name":"Bash","tool_input":{"command":"echo f3"},"tool_response":{"stdout":"f3"}}' "$R/wsA" \
  | base hook post-tool-use
```

## Evidence

```
[...]

Issue #76, opened 2026-09-07
`base hook stop` prints the relay task nudge as plain stdout, which Claude Code never delivers to the model on Stop


ChristopherKahler:
## Summary

`base hook stop` prints the relay task nudge (`relay_task_tick(..., Phase::Stop, ...)`) as plain stdout. Claude Code does not deliver plain stdout to the model on `Stop`, in that turn or the next. The nudge is recorded as hook output and never read by the model.

## Impact

An autonomous run that finishes a turn with an open relayed task gets no end-of-turn reminder. The relay layer believes it nudged (the block was printed, the hook exited 0), the operator's board shows the task as delivered, and the session idles. This is the "silent no-op that reads as a clean pass" shape.

## Mechanism (v0.14.1 = `972996a`)

- `src/hook/mod.rs`, `run()`, arm `"stop"`: after `stop::handle(&config, &cwd)?`, `relay_task_tick(sid, &cwd, &config.relay, Phase::Stop, false)` returns a block and the arm does `print!("{block}")`.
- Claude Code's hooks reference, "Exit code 0": "For most events, Claude Code writes stdout to the debug log and doesn't show it in the transcript. The exceptions are `UserPromptSubmit`, `UserPromptExpansion`, `SessionStart`, and `PostModelSwitch`, where Claude Code adds plain-text stdout as context that Claude can see and act on." `Stop` is not in the list. The only model-facing outputs on `Stop` are `decision: "block"` + `reason` (forces a continuation) and `systemMessage` (user-visible only).
- base already encodes this rule for one event: `src/hook/pre_tool_use.rs:14-17` and the `"pre-tool-use"` arm in `mod.rs` wrap PreToolUse context in `hookSpecificOutput.additionalContext` because "plain stdout is transcript-only on this event". The `stop` arm did not get the same treatment.

## Reproduction (run 2026-09-07 15:33 local, Windows 10)

Binary: `C:\Users\Chris\.local\bin\base.exe`, `base --version` = `base 0.14.1`. Claude Code `2.1.263`, model `claude-haiku-4-5-20251001`. The probe replaces base's hook with a nonce-printing hook so the channel is tested in isolation; base's own code path is the `print!` cited above.

`hook.py` (writes a `fired.log` line per invocation, then prints per arm):

```python
import sys, json, datetime
arm = sys.argv[1]; n = "1788813182"
raw = sys.stdin.read()
try: ev = json.loads(raw).get("hook_event_name")
[...]

PR #102, opened 2026-09-08
fix(hook,relay): four silent no-ops - post-tool context, log tier, ping slug, line endings

### LINKED ISSUE #75: `base hook post-tool-use` prints its context as plain stdout, which Claude Code never delivers to the model on PostToolUse (section AST context and extension nudges are silent no-ops)
## Summary

`base hook post-tool-use` writes its two model-facing injections (section AST context on a partial `Read`, extension `inject` nudges on `Write`/`Edit`) with `print!` as plain stdout. Claude Code does not deliver plain stdout to the model on `PostToolUse`; only `hookSpecificOutput.additionalContext` reaches it. The hook exits 0, `hook-events.jsonl` records `section_context: true` / `nudged: true`, and the model never sees the text.

## Impact

Every partial read of a mapped source file since the section-context feature landed has produced a call-edge block that no model has read. Every extension `inject` nudge (the `[[hooks.post_tool.handlers]]` surface plugins are told to use) has been a no-op. The hook log says both fired, which is the worst shape: a silent no-op that reads as a clean pass. base already knows this rule and applies it to `PreToolUse` only (`pre_tool_use.rs:14-17`), which is why the pre-tool `[AST] … N entities` line does reach the model and the post-tool `[AST] Lines a-b of …` block does not.

## Mechanism (v0.14.1 = `972996a`)

- `src/hook/post_tool_use.rs:96-108`: on `tool_name == "Read"` with `offset` and `limit`, `crud::ast_query::section_entities(...)` is printed with `print!("{}", section.trim_end())` and `data.section_context = true`.
- `src/hook/post_tool_use.rs:171-195`: extension `inject` handlers print `message` with `print!("{}", message.trim_end())` and set `data.nudged = true`.
- `src/hook/mod.rs`, arm `"post-tool-use"`: no envelope is built; contrast the `"pre-tool-use"` arm, which wraps all pre-tool text in `{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":…}}` with the comment "PreToolUse context only reaches the model through the JSON envelope — plain stdout is transcript-only on this event."
- Claude Code hooks reference, "Exit code 0": "For most events, Claude Code writes stdout to the debug log and doesn't show it in the transcript. The exceptions are `UserPromptSubmit`, `UserPromptExpansion`, `SessionStart`, and `PostModelSwitch`, where Claude Code adds plain-text stdout as context that Claude can see and act on." PostToolUse decision control lists `additionalContext` as the context field.

## Reproduction A: base's own path (run 2026-09-07 15:36 local, Windows 10)

Binary `C:\Users\Chris\.local\bin\base.exe`, `base --version` = `base 0.14.1`. Claude Code `2.1.263`, model `claude-haiku-4-5-20251001`, the operator's live `~/.claude/settings.json` (base hooks wired as installed).

Scratch app with one Python file and a fresh map:

```
mkdir -p f1app/src && cd f1app && git init -q .
# src/probe_mod.py: alpha_one, alpha_two (calls alpha_one), class Beta.gamma (calls alpha_two), delta_last
base sync --ast --target "$PWD" --yes          # .base-ast/ast.ttl, 6 entities
base ast query --file src/probe_mod.py         # sanity: 6 entities listed
claude -p "Use the Read tool on src/probe_mod.py with offset 3 and limit 5. Then reply with exactly two lines: line 1 the word done; line 2 any text you received that begins with [AST] or that was not part of the file's own content, or the word NONE." \
  --model haiku --allowedTools Read --permission-mode acceptEdits --output-format json
# session 0fbec36b-bff3-41ac-9ac0-ec6b623a7903
```

### Evidence A

Model reply (it saw the PreToolUse map line and nothing from PostToolUse):

```
done
[AST] C:\...\scratchpad\f1app\src\probe_mod.py - 6 entities
```

base's own log, `C:/Users/Chris/.base/hook-events.jsonl`, the line for this Read:

```
"hook":"post-tool-use" "tool_name":"Read" "section_context":true "success":true "ts":"2026-09-07T15:36:30-05:00" "session_id":"0fbec36b-bff3-41ac-9ac0-ec6b623a7903"
```

Transcript `~/.claude/projects/.../0fbec36b-bff3-41ac-9ac0-ec6b623a7903.jsonl` (29 lines). Every `hook_additional_context` attachment (the model-visible channel):

```
L5  hookEvent=SessionStart  hookName=SessionStart      (context-mode plugin text)
L17 hookEvent=PreToolUse    hookName=PreToolUse:Read   (context-mode plugin text)
L19 hookEvent=PreToolUse    hookName=PreToolUse:Read   content=["[AST] C:\...\probe_mod.py - 6 entities"]   <- base pre-tool envelope, delivered
```

Every `hook_success` record:

```
L4  SessionStart:startup   command="base hook session-start"     content_len=2447
L11 UserPromptSubmit       command="base hook user-prompt-submit" content_len=8253
L18 PreToolUse:Read        command="base hook pre-tool-use"      content_len=13    (stdout was the JSON envelope)
L21 PostToolUse:Read       command="base hook post-tool-use"     content_len=346   <- printed, captured, not attached
```

L21's `content` is base's section block, verbatim:

```
[AST] Lines 3-8 of C:\...\f1app\src\probe_mod.py:
  fn alpha_one() (line 3)
    called_by: src\probe_mod.py → alpha_two()
  fn alpha_two() (line 6)
    calls: alpha_one()
    called_by: src\probe_mod.py → .gamma()
```

There is no `hook_additional_context` with `hookEvent=PostToolUse` in the transcript and the model's reply does not contain "Lines 3-8".

## Reproduction B: the channel in isolation (run 2026-09-07 15:33 local)

Same Claude Code and model; a nonce hook replaces base so the event's two channels are compared side by side in one session. `hook.py` appends a `fired.log` line per invocation and prints per arm: `ups` → `NONCE_UPS_1788813182` (plain, UserPromptSubmit, the control), `ptu` → `NONCE_PTU_PLAIN_1788813182` (plain, PostToolUse), `ptuj` → `{"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"NONCE_PTU_JSON_1788813182"}}` (JSON, PostToolUse, the fix shape), `stop` → `NONCE_STOP_1788813182`. `settings.json` wires `ups` on UserPromptSubmit, `ptu` and `ptuj` in one PostToolUse group, `stop` on Stop; passed with `--settings`.

```
claude -p "Use the Bash tool to run exactly: echo probe. Then reply with one line listing every string containing the word NONCE that you can see anywhere in your context, or the word NONE." \
  --model haiku --allowedTools Bash --permission-mode acceptEdits --settings settings.json --output-format json
# session f61bfc8a-225f-4dff-adaf-f419d7b15b9f
```

### Evidence B

`fired.log`: `arm=ups event=UserPromptSubmit`, `arm=ptu event=PostToolUse`, `arm=ptuj event=PostToolUse`, `arm=stop event=Stop` (all four executed).

Model reply: `NONCE_UPS_1788813182, NONCE_PTU_JSON_1788813182`.

Transcript lines carrying each nonce:

```
NONCE_UPS         L11 hook_success(UserPromptSubmit)   L23 assistant message
NONCE_PTU_PLAIN   L18 hook_success(PostToolUse:Bash)   (nothing else)
NONCE_PTU_JSON    L19 hook_success  L20 hook_additional_context(PostToolUse:Bash)  L23 assistant message
```

Two hooks, same event, same tool call, same session: the JSON one is attached and read, the plain one is captured and dropped.

## Fix shape

The `"post-tool-use"` arm in `mod.rs` assembles handler output into the same envelope the `"pre-tool-use"` arm uses, with `hookEventName: "PostToolUse"`. `post_tool_use::handle` returns `(HookEventData, String)` like `pre_tool_use::handle` instead of printing. Stderr diagnostics stay on stderr. Operator-only lines (none today on this event) would go to `systemMessage`. This is the `claude` emitter row for post-tool-use in `base-harness-compat-spec.md` §5.6; the Stop half is the sibling issue.

## Red-first test

`tests/hook_post_tool_channel_test.rs`: temp app with a two-function source file and a built map; spawn `base hook post-tool-use` with a `Read` payload carrying `offset`/`limit`; assert stdout parses as JSON with `hookSpecificOutput.hookEventName == "PostToolUse"` and `additionalContext` containing `[AST] Lines`, and that no non-JSON text precedes it. Fails on 0.14.1 (stdout is the plain block).

Related: base-harness-compat-spec.md §2.4 F1, §5.6.

ChristopherKahler:
Closes #75, #76, #77, #86, #95.

Four silent no-ops in the hook and relay layers: code that runs, returns success, and delivers nothing. Each one is fixed at the mechanism, not at the call site the issue happened to name.

## What changed

**1 — PostToolUse context reaches the model.** The post-tool-use arm now returns its text and `mod.rs` wraps it in the `hookSpecificOutput` envelope with `"hookEventName": "PostToolUse"`, where the host actually reads it. All **four** print sites moved, not the two the issue named — `emit_move_nudge` (`post_tool_use.rs:407`, called at `:28`) was one of the two extra. `session-start` and `user-prompt-submit` keep plain stdout, because that is where the host delivers it for those events. Stop stops writing into a dead channel and emits `systemMessage` instead.

**2 — The hook event log lands in the tier the payload named.** `log_hook_event` takes `payload_cwd` and `cwd_from_payload` (`mod.rs:433`), so a hook fired for workspace A no longer logs into the process cwd's workspace B. Fixed on the **failing** arm too, via `struct HookRun` + `fn run_event` — the original only handled the success path. Absent cwd is distinguished from empty by a `"cwd_source"` field, and `nudged` / `standards_injected` now reach the log line.

**3 — Ping slugs cannot collide.** `relay::ping_slug()` (`relay/mod.rs:599`, called at `cli.rs:2727`) replaces the old scheme where 8 pings produced 2 files and 6 were silently overwritten.

**4 — Scripts check out runnable on Windows.** `.gitattributes` pins `.sh`/`.py` to LF and `.ps1` to CRLF, enforced by `scripts/test-crlf-attributes.sh` as a blocking CI step.

7 commits, 12 files, 936 insertions. Rebased twice onto moving main (`3e65bcac`, then `4ee4a45`); both conflicts were the `scripts/changelog-notes/0.14.2.md` add/add tail only, resolved by ORDERING — main's paragraphs keep their order, mine append last, and the resolver asserts no paragraph from either side is lost and no conflict marker reaches the file.

## Design-to-code map

| lock | promise | symbol at file:line | pinned by |
|---|---|---|---|
| 1 | PostToolUse context travels in the envelope | `mod.rs` post-tool-use arm, `"hookEventName": "PostToolUse"` | `hook_output_channel_test.rs::post_tool_use_context_travels_in_the_json_envelope`; harness #75 |
| 1 | the handler returns its text instead of printing | `post_tool_use.rs::handle -> Result<(HookEventData, String)>` | compiles; 0 `print!`/`println!` left in the file, asserted by `rebase_verify.sh` |
| 1 | **all four** print sites, not the two the issue named | `emit_move_nudge -> String` (`post_tool_use.rs:407`), called at `:28` | harness #75 drives the `path-drift` block specifically |
| 1 | plain stdout KEPT where the host delivers it | `session-start` and `user-prompt-submit` arms | `the_events_that_deliver_plain_stdout_still_use_it` |
| 1 | Stop stops writing into a dead channel | `mod.rs` stop arm, `json!({ "systemMessage": … })` | `stop_never_writes_a_bare_block_to_stdout`; harness #76 row 2 |
| 2 | the log tier comes from the payload | `log_hook_event(…, payload_cwd, cwd_from_payload)` (`mod.rs:433`) | `hook_log_tier_test::the_line_lands_in_the_workspace_the_payload_named`; harness #77 |
| 2 | …on the FAILING arm too | `struct HookRun` + `fn run_event` | `a_failing_hook_still_logs_into_the_payloads_workspace` |
| 2 | absent ≠ empty | `"cwd_source"` in the event JSON | `a_payload_with_no_cwd_says_so…`; harness #77 |
| 2 | `nudged` and `standards_injected` reach the log | same JSON builder | `the_log_line_carries_every_field_the_struct_sets`; harness #77 |
| 3 | a ping slug that cannot collide | `relay::ping_slug()` (`relay/mod.rs:599`), called at `cli.rs:2727` | `relay_ping_slug_test`, 4 tests; harness #86 |
| 4 | scripts check out runnable on Windows | `.gitattributes` | `scripts/test-crlf-attributes.sh`, a blocking CI step |

No empty cells.

## Verification on this head (`f91fe29`)

| stage | result |
|---|---|
| release build | GREEN, 1m16s |
| `cargo clippy --all-targets -- -D warnings` | GREEN |
| `cargo test` | GREEN, every target, 0 failed |
| `verification/base-0.14.2/silent_no_ops_0142.sh` | **12 PASS / 0 FAIL / 0 SKIP** |

### Harness contract

```
BASE_BIN=<branch binary>  OLD_BIN=~/.cache/merlin-audit/v0.14.1/base  REPO_DIR=<worktree>
```

Provenance is asserted before any row runs: the branch binary aside carries `ping_slug`, md5 `fee995a57dcfebb2ad97118eef09ebe3`; the v0.14.1 control is md5 `da7950677bce8059247e7dd858f9016b` and does **not** carry the symbol. The harness aborts if the two md5s match, or if the control carries the symbol — a run against two copies of the same binary cannot report a pass.

### Controls asserted RED in the same run

```
#75 old: bare stdout, no envelope          (RED as expected)
#77 old: line lands in B, the process cwd  (RED as expected)
#86 old: 8 pings -> 2 files, 6 replaced    (RED as expected)
#86 new: 8 pings -> 8 files
#76 new: 1 open task -> systemMessage envelope, no bare block
```

## What the build caught that reading had not

1. **`post_tool_use.rs` early return** still returned the old shape, and it sits *after* the move nudge is appended. `Ok((data, String::new()))` compiles and silently drops the block this PR exists to deliver.
2. **Three `needless_borrow`.** `run_event` takes `&Value`, so each surviving `&stdin_json` was a double borrow. Deref coercion kept `cargo build` green; only `clippy -D warnings` found them.
3. **The discriminator masked its own counter.** `((pid & 0xff) << 8) | (seq & 0xff)` capped the counter at 8 bits and wrapped every 256 calls: 10,000 in-process pings gave 1,280 distinct slugs, 8,720 collided. Caught by a test written for this PR.
4. **The harness refused a correct binary.** `strings BIN | grep -q SYM` under `pipefail` reports the pipeline FAILED when it MATCHES — grep exits at the first hit and `strings` takes SIGPIPE. Fixed by dumping to a file.
5. **The #76 row was vacuous.** It ran `hook stop` with no open task, so the `systemMessage` branch never executed. It now builds a real open task with the product itself and fails if Stop says nothing.

## Known and deliberately not fixed

`silent_no_ops_0142.sh` prints two `No such file or directory` lines from `wc -l` reading a deliberately absent `hook-events.jsonl` in the #77 section; shell redirection failures bypass the command's own `2>/dev/null`. The PASSes are sound — the arithmetic was checked in both directions: the old arm needs A=0/B=1 and the new arm A=1/B=0, and the missing file yields the 0 each one requires. Cosmetic log noise, no verdict depends on it.

---

https://claude.ai/code/session_016XvBnm3p7R8utD61wL4sxP


ChristopherKahler:
## Known defect in this PR's own acceptance harness — provenance block, one-sided

Filing this against the code rather than leaving it in review chatter. It is **not** fixed on this branch, deliberately; see the last section.

### Where it is

`verification/base-0.14.2/silent_no_ops_0142.sh`, the provenance block:

```sh
strings "$BASE_BIN" > "$WORK_SYMS_NEW" 2>/dev/null || true
strings "$OLD_BIN"  > "$WORK_SYMS_OLD" 2>/dev/null || true
if ! grep -q "$SYMBOL" "$WORK_SYMS_NEW"; then
    echo "ABORT: branch binary has no '$SYMBOL' — it is not this branch's build."; exit 2
fi
if grep -q "$SYMBOL" "$WORK_SYMS_OLD"; then
    echo "ABORT: the CONTROL carries '$SYMBOL' — it is not 0.14.1."; exit 2
fi
echo "  symbol '$SYMBOL': present in branch, absent in control — both binaries are what they claim"
```

### The defect

Both `strings` calls end in `|| true`, so a failed dump is indistinguishable from a successful one that found nothing. The two checks then treat that empty file differently:

- **Branch dump empty** → `! grep -q` is true → **ABORT**. Fails closed. Correct.
- **Control dump empty** → `grep -q` is false → **no abort**, and the script goes on to print *"present in branch, absent in control — both binaries are what they claim."* At that point the sentence is false, and it is the one line the block exists to produce.

A zero from an absent symbol and a zero from a dump that never got written are the same number. Only one of them is evidence.

### Scope, stated precisely

The hole is **one-sided**. It cannot green a wrong *branch* binary — that path aborts. It can only make a printed provenance claim false about the *control*. The four locks are proven by the rows themselves, not by this block, so no row's PASS/FAIL is affected. What is affected is the reader's warrant for believing the run was against the two binaries it names.

### The fix

Assert both dumps are non-trivially sized before trusting either `grep`, e.g.:

```sh
for f in "$WORK_SYMS_NEW" "$WORK_SYMS_OLD"; do
    n=$(wc -l < "$f")
    [ "$n" -ge 1000 ] || { echo "ABORT: symbol dump $f has $n lines — strings failed, counts are void."; exit 2; }
done
```

Threshold is a floor for "a real Rust binary's symbol table", not a tuned number; both binaries here dump far above it.

### Why it is not fixed in this PR

A commit moves the head sha, which voids the CI run already read at step level on `f91fe29` and forces push, CI and the independent G2 rerun from scratch — to repair a sentence that cannot change any verdict. It lands as a follow-up once this is merged.

### Credit

Found by the orchestrator adding a generic guard-the-guard to their own independent harness — assert the dumps are non-empty before believing a zero — with no defect suspected. It surfaced a live one in a shipped instrument.

---

https://claude.ai/code/session_016XvBnm3p7R8utD61wL4sxP

ChristopherKahler (commented):
**Second-builder review (0.14.2 merge rule, line 6)** — plover, on head `f91fe29adb088c14eed690cf002aed536147bf39`.

## Verdict: APPROVE the change. One instrument defect must be fixed before this harness is re-run at the release gate.

A GitHub `APPROVED` state is unobtainable on this repo (the API refuses self-approval with 422 since Chris authors every PR), so the verdict is stated here in the body, as the merge rule provides for.

Reviewed the delta at head — 12 files, +936/−39 — through the API rather than a worktree, because WSL is quiesced for viper's build window.

---

## Findings

### F1 · the provenance block cannot run on Windows Git Bash, and it blames the binary — VOID FAIL

`verification/base-0.14.2/silent_no_ops_0142.sh`. **`strings` is not on PATH in Windows Git Bash.** Measured on this machine:

```
$ command -v strings
NO strings ON PATH
$ strings /c/Users/Chris/.local/bin/base > "$T" 2>/dev/null || true
$ wc -c < "$T"
0
  -> harness WOULD ABORT: 'branch binary has no ping_slug'
```

The `|| true` swallows *command not found*, the dump is empty, and the next line aborts with

> `ABORT: branch binary has no 'ping_slug' — it is not this branch's build.`

**The tool's absence and a genuine provenance failure share an exit path and a message that names the wrong subject.** That is law 24 exactly — the false-FAIL twin the round already paid for once with `cargo-clippy.exe` not installed on the MSVC toolchain — and law 21's "a harness line is only as true as its label."

**It fails in the other polarity in the same block.** Two lines down:

```sh
if grep -q "$SYMBOL" "$WORK_SYMS_OLD"; then
    echo "ABORT: the CONTROL carries '$SYMBOL' — it is not 0.14.1."; exit 2
fi
```

With an empty dump this check **passes vacuously, always** — it is a control-side provenance assertion that on this machine can only ever succeed, having read zero bytes. That is law 23's false-PASS. So the same block produces a void FAIL on one arm and a guaranteed PASS on the other, and neither is about the binary.

This is not hypothetical for the gate: the release gate re-runs the ledger regression rows, and anything driven from the Windows side hits this.

**Fix** — law 24's own prescription, assert the tool before believing either result:

```sh
command -v strings >/dev/null 2>&1 || {
    echo "ABORT: 'strings' unavailable — provenance cannot be established, and neither"
    echo "       symbol check below would mean anything. This is a toolchain gap, NOT"
    echo "       a defect in either binary."; exit 2; }
```

### F2 · a SKIP on a control arm exits 0

The tail is correct and I want to say so explicitly, because the round has been sweeping for the opposite:

```sh
[ "$fail" -eq 0 ] || exit 1
```

That tests **the job's own counter**, which is the assigned-not-inherited shape law 27 asks for — not a trailing `grep`/`echo` whose subject is something else. No law-27 defect here.

But `$skip` is printed and never gated, and three legs can skip:

- `#75` — `skipped "$gen: the move nudge did not fire — row measured nothing"`, reachable on **either** generation
- `#86` — `skipped "old: ... the race did not fire this run, so the control proved nothing"`
- `#95` — skipped whenever `REPO_DIR` is unset, which is the default

When the **old** arm skips, the control is void and the new arm's PASS proves strictly less — and the exit code cannot say so. A consumer reading only the status cannot separate "all green" from "green with a dead control". That is law 24 one level up from the individual row.

**Fix:** count control-arm skips separately and exit non-zero on any, or at minimum exit `3` on `skip > 0` so the status distinguishes the two states.

### F3 · the EXIT trap is replaced, so the two `strings` dumps leak — verified, minor

```sh
WORK_SYMS_NEW=$(mktemp); WORK_SYMS_OLD=$(mktemp)
trap 'rm -f "$WORK_SYMS_NEW" "$WORK_SYMS_OLD"' EXIT
...
WORK=$(mktemp -d); trap 'rm -rf "$WORK"' EXIT
```

Bash keeps one EXIT handler; the second `trap` replaces the first. Verified rather than asserted:

```
--- after exit ---
  LEAKED (still exists): /tmp/tmp.9Cur9DK2E2
  LEAKED (still exists): /tmp/tmp.AtKJB5C26E
  cleaned: /tmp/tmp.9F5T8CprOP
```

Two full binary dumps per run, left behind. **Fix:** one trap naming all three paths.

---

## What I verified rather than took on trust

- **The stdout-channel claim is exactly true at head.** Mapped every stdout site in `src/hook/mod.rs` to its arm at `f91fe29`: bare `print!` survives **only** on `session-start` (155, 162) and `user-prompt-submit` (232, 238) — the two events where the host does deliver plain stdout — while `pre-tool-use` (197), `post-tool-use` (220) and `stop` (258) each emit a JSON envelope. That matches the changelog paragraph word for word.
- **No `print!` survives in `post_tool_use.rs`.** At head it holds four `eprintln!` and nothing else, and the handler's doc comment scopes stderr deliberately: *"those lines are operator diagnostics, not model context, and they were never on the broken channel."* Checked the whole file, not the diff — a patch cannot show an untouched `print!`.
- **The `post-tool-use` arm runs no relay tick**, so there is exactly one JSON object on stdout for that event.
- **The one path that would have silently dropped the fix is called out and handled**: a Bash `mv` carries no `file_path`, so the early return is the only exit that case reaches, and it returns `(data, output)` with a comment saying an empty string there *"would compile and silently drop it."* That is the law-25 shape — the leg that separates repaired from silenced — reasoned at the source rather than only in a test.

## Craft worth recording

- `ping_slug` carries **both** discriminators with a stated reason neither suffices alone (a counter cannot separate two processes; a pid cannot separate two pings from one process), and the doc comment records a regression the author caught in their own fix: masking the counter to 8 bits wrapped every 256 calls, so 10,000 pings yielded 1,280 distinct slugs. *"Tidiness is not worth the property."*
- The harness explicitly avoids the `strings | grep -q` + `pipefail` SIGPIPE inversion — dumping to a file first, with the reason in a comment.
- `#77`'s control is two-directional: new asserts A=1/B=0, old asserts A=0/B=1.
- `#76`'s empty-path row is paired with a real open task built **by the product itself**, and `pending == 0` is a `bad`, not a `skipped` — the author refused to let a vacuous leg count.

## Merge position

F1–F3 are all in the acceptance harness, not the shipped code. The product changes are additionally covered by three new `cargo test` files, and cargo produces a real exit status with no shell wrapper to invert or swallow — the reason cargo-native work escaped the round's runner sweep.

**Nothing here blocks the merge.** F1 must land before this harness is re-run at the release gate, or the run will abort blaming a good binary on any Windows-side invocation.

PR #103, opened 2026-09-08
fix(install): write the binary under every name Windows needs (#91)

### LINKED ISSUE #91: Windows: `base install` writes ~/.local/bin/base with no .exe, so a fresh install cannot be run by name
**Version:** 0.14.1
**Host:** Windows 10 Pro 19045

## What happens

`base install` copies the binary to `~/.local/bin/base` with no extension (`src/install.rs:1605`, one path shared with Unix). On Windows that file is a valid PE image with a name Windows cannot resolve, because PATHEXT requires an extension.

A fresh install reports success and prints correct PATH guidance. Then `base` does nothing.

## Reproduce

Install into a home that has never had base, and put only that install on PATH:

```
PATH = <fresh>\.local\bin ; C:\Windows\System32

cmd.exe > base --version
  'base' is not recognized as an internal or external command,
  operable program or batch file.
  exit 1

pwsh    > base --version
  (no output)
  exit 0
```

The file itself is fine:

```
bin dir:  base   25,297,408 bytes
magic:    MZ                      <- valid Windows PE
```

Renaming it to `base.exe` makes both shells return `base 0.14.1` at exit 0.

## Why it has not been noticed

An existing Windows install has **both** files: `base` from `base install`, and `base.exe` from the update path, which does write the extension. `base.exe` is what actually runs, so the extensionless copy looks like harmless clutter rather than the whole install on a clean machine.

This only bites the path a new user takes, which is the path the one-line installer creates.

## Suggested fix

On Windows, `install::run` should write `base.exe`. `install.ps1` currently copies `base` to `base.exe` after the handoff as a workaround; that belongs in the binary.

Worth a test asserting the installed filename ends in `.exe` under `cfg(windows)`.

ChristopherKahler:
Fixes the clean-machine Windows install. `install.rs` wrote `~/.local/bin/base` with no extension, so PATHEXT resolves neither cmd nor PowerShell `base` to it. Existing installs hid this because they already carry both names.

## The fix

`home::base_binary_names()` returns every name, primary first: both on Windows, one on Unix. Install writes all of them, uninstall removes all of them.

Renaming the destination instead would have regressed Git Bash. `refresh_sibling` (`update/mod.rs:367`) documents both names as the intended Windows layout and exists to keep them in step, but it returns early when the sibling is absent (`:374`), so it refreshes a pair and never creates one. Install was the only thing that could make the pair, and it made half.

`manifest.rs` moved with it, because it recorded the literal while detecting the literal, and `manifest.rs:317` documents a stale path pinning the install at the old version forever.

## The five commits

```
f743d35 test(guard): the #91 guard prints what it scanned, and zero files is a failure
8dee665 docs(changelog): the 0.14.2 note for the Windows exe-suffix fix (#91)
66607a3 fix(install): route every binary-name site through the seam, and teach the guard the shape it could not see
232fb01 fix(install): uninstall reports the directory and every name it looked for
f42e4e1 fix(install): write the binary under every name Windows needs (#91)
```

`232fb01` and `f42e4e1` are the original two, replayed by the rebase onto `36ca97b`.

### What the last three added

**Double tick.** `install_binary` ended with the step's `✓` and is called once per name, so Windows printed two ticks under one heading. The tick moved to the caller.

**`install_dest()` now calls `base_binary_path()`** — a proven-identical substitution that also gives that function its only caller.

**Every binary-name site routed through the seam, zero exemptions added:** `update/mod.rs` 156, 300, 607, 694, 704. Lines 616 and 682 are genuine platform branches and were deliberately left alone.

Site 156 is the one worth re-reading, because `find_base_binary` searches an *extracted archive* and that name need not be the *installed* name. `.github/workflows/release.yml` settles it: the matrix sets `binary: base.exe` for `x86_64-pc-windows-msvc` and `base` for the three Unix targets, both packaging steps copy that one member, and `asset_name()` maps `(OS, ARCH)` to a single asset — so the archive searched is always the archive built for the running platform, and its member is `base` + `EXE_SUFFIX` by definition. What would have been wrong is the *plural*: an archive holds one member, a Windows install holds two. `base_binary_names()` is an install-layout fact with no meaning inside a zip.

**The guard learned the `cfg!(windows)` name-arm** — `contains("cfg!(windows)") && contains("\"base.exe\"") && contains("\"base\"")`, matched on one line, the ternary form. A multi-line `if cfg!(windows) {` block is deliberately not matched.

## Guard evidence

Red-then-green through cargo itself, not through a side harness:

| Arm | Tree | rc | Output |
| --- | --- | --- | --- |
| RED | pre-fix `update/mod.rs` | 101 | names exactly 156, 300, 607, 694, 704 — nothing else |
| GREEN | fixed | 0 | `5 passed; 0 failed; 0 ignored; 0 filtered out` |

`f743d35` adds a visited-count assert to `no_literal_base_binary_path_outside_the_seam`, because a guard that walks a set must print the size of the set it walked. If `rust_files` returns empty — `src/` not where the test is standing — every `continue` is taken, `offenders` stays empty, `is_empty()` is true, and the guard goes green having opened no files. That is the exact shape that produced a false PASS on an isolation tripwire earlier in this release.

Three arms, because an assert that has never fired is untested code that always passes:

| Arm | rc | Output |
| --- | --- | --- |
| A control — scan root pointed at a nonexistent dir | 101 | `scanned 0`, `opened ZERO files … proved NOTHING` |
| B red — pre-fix `update/mod.rs` | 101 | `scanned 98`, five sites named |
| C green — clean | 0 | `scanned 98`, `5 passed; 0 failed; 0 filtered out` |

Arm B's non-zero count is the half that matters: a firing visited-assert would have *masked* the offender list and graded the wrong failure.

## Known and deliberately deferred to 0.14.3

`guard_test.rs` holds three scanning guards. **One of the three now prints a visited count and asserts it is non-zero** — `cargo test --test guard_test -- --nocapture` on this branch emits exactly one `guard: scanned …` line, and that single line is the gap made visible. `no_raw_home_dir_outside_the_seam` and `no_hand_built_sparql_change_outside_the_seam` are missing the same instrument.

Their **exposure** is not identical, and the difference is worth stating rather than flattening. Both siblings walk `Path::new(env!("CARGO_MANIFEST_DIR")).join("src")` — an absolute path fixed at compile time. The hardened guard walked `Path::new("src")`, a **relative** path resolved off the process working directory, which is precisely why it was the one that could silently walk an empty set. Same missing assert; lower odds of firing.

So the deferral is a scoped orchestrator decision, not an oversight: the guard whose walk could actually miss is the guard that got the assert, and the two whose roots are compile-time absolute are queued behind it. A reviewer should read the omission as bounded rather than missed.

## Verification

| Instrument | Result |
| --- | --- |
| cargo test | 835 passed / 0 failed |
| seam guard, merged tree | rc 0 — `5 passed; 0 failed; 0 filtered out`, `scanned 98` under `--nocapture` |
| cargo clippy | rc 0, 0 errors, 0 warnings, on the pre-rebase head |
| Windows acceptance leg | 6/6 PASS, script exit 0 |
| build | base.exe 25,163,776 b, md5 0A9257BE61CF3B79763A4E2E70BC58E4 |

The Windows leg invokes the binary directly. `install.ps1` is absent from the run by design: the installer write-alongside workaround is exactly what makes a broken binary look correct, so a leg routed through it could only ever prove the workaround works.

`Cargo.toml` reads 0.14.1 on this branch and `scripts/release.sh` owns the bump, so a binary self-reporting 0.14.1 is correct, not a stale build.

## The rebase onto 36ca97b

#106 deleted 1046 bytes mid-file from `scripts/changelog-notes/0.14.2.md`, the same file `8dee665` appends to, so the two sides now touch one common path. That was measured rather than assumed.

The post-rebase blob was predicted at 18,069 bytes / CR=0 / LF=41 (17,212 on main + an 857-byte append) *before* rebasing, and came out at exactly that. `git diff --numstat` against main is still `2 0` on the file, with zero conflict markers and no 3+ newline runs. Line 9 of the merged tree is byte-identical to main's deduplicated version (2,244 → 1,198 bytes), so #106's fix survives intact and the append sits after it. `git diff --numstat` between the pre- and post-rebase heads is that one changelog line: the code tree is unchanged across the rebase.

## Not covered

No Linux acceptance leg. `EXE_SUFFIX` is empty there so the behaviour is unchanged by construction, but that is an argument, not a measurement.

Closes #91

https://claude.ai/code/session_0131aPBXBkbdHTtin3MDLVLq


ChristopherKahler (commented):
Second-builder review (0.14.2 merge rule, line 6) — merlin, on head `e122dac95e41d80c3f5c44f7124b408a3a12bbbb`.

## Verdict

**The #91 fix is correct and I would merge it.** `base_binary_names()` writes both Windows names on a clean install, uninstall removes both, and the manifest records the name that exists. The central claim holds: a clean-machine Windows install now produces a `base.exe` that PATHEXT can resolve, and the extensionless `base` that Git Bash needs survives alongside it.

Four findings below. **F1 is a regression this PR introduces in the command it fixes and should be fixed here.** F2 is the one that matters for durability and is fine as a follow-up. F3 and F4 are small.

## What I verified

- Read the full diff (4 files, +159/−17) plus `install.rs`, `manifest.rs`, `home.rs` and `update/mod.rs` in full at the head sha.
- Enumerated **every** `base_binary_*` reference across all **156** `.rs` files in the tree at head, not a sample.
- CI run on this sha: conclusion `success`, and green at **step** level — `cargo test` ✓, `release rehearsal` ✓, `test-release-invocations.sh` ✓, `test-release-packaging.sh` ✓, and `cargo clippy --all-targets -- -D warnings` ✓. The clippy job is `continue-on-error: true` (`ci.yml:59`), so its *job* conclusion is green regardless; the *step* conclusion is what I read, and it passed.
- `test-crlf-attributes.sh` is absent from this run's `release-invocations` job. That is correct, not a skip: the script arrives with PR 102, which is not merged, and this branch is based on `main` @ `4ee4a455`.

No self-delete hazard from the two-name loop: on Windows, `base install` running from `base.exe` hits `already_in_place` on the first destination and copies to the extensionless `base` second, so `remove_file` never targets the running image. The gated-installer path described at `install.rs:347` behaves the same way with the names reversed.

## F1 — install output is malformed, on every platform, and prints `✓` twice on Windows

`install.rs:40` announces the step with `println!`, and `install_binary` still ends with `println!("✓")` at `install.rs:373` — now called once per name.

```
1. Install binary → C:\Users\x\.local\bin\base.exe, C:\Users\x\.local\bin\base ...
✓
✓
```

Every other step in this installer uses `print!("N. … ")` then `println!("✓")` on one line. Moving the heading to `println!` breaks that idiom on Unix too, where the output is now a heading line followed by a lone `✓`.

The comment at `install.rs:338-339` anticipated the duplicated *heading* and moved it out. The duplicated `✓` came along with it.

Fix is one line each: make the caller `print!` rather than `println!`, drop `println!("✓")` from `install_binary`, and print it once after the loop.

## F2 — the seam is incomplete, and the guard cannot see the shape that caused #91

The doc comment on `base_binary_name()` (`home.rs:195`, closing at `home.rs:194`) names the cause exactly: *"A `cfg(windows)` arm at each call site is what produced #91 … One expression cannot drift, and `tests/guard_test.rs` keeps a sixth site from appearing."*

Two of those arms are still in the tree at head, untouched by this PR:

```rust
// update/mod.rs:156, in find_base_binary
let target = if cfg!(windows) { "base.exe" } else { "base" };

// update/mod.rs:300, in install_dest
let name = if cfg!(windows) { "base.exe" } else { "base" };
```

The guard tests two substrings — `"~/.local/bin/base"` and `.join("base")` (`guard_test.rs:186-187`). Neither matches `if cfg!(windows) { "base.exe" } else { "base" }`. So the guard is blind to the precise expression form that produced sites one through five, and two live instances of it remain. "Keeps a sixth site from appearing" is true for the literal shape and not for the cfg-arm shape.

Sharpest version of this: `install_dest()` at `update/mod.rs:299-303` is `home::base_binary_path()` rewritten by hand.

```rust
// update/mod.rs — live, called
crate::home::home_root().map(|h| h.join(".local").join("bin").join(name))

// home.rs:218-219 — added by this PR
Some(home_root()?.join(".local").join("bin").join(base_binary_name()))
```

Same path, same semantics, two expressions. This is the drift the PR exists to end, reproduced inside the PR that ends it.

Not a correctness bug today — `cfg!(windows)` yields the same string `base_binary_name()` does. It is a completeness gap plus a guard that reports coverage it does not have.

Suggested: point `install_dest()` at `base_binary_path()` and `find_base_binary`'s `target` at `base_binary_name()`, then extend the guard to flag `cfg!(windows)` / `cfg(windows)` lines mentioning `base.exe` outside `home.rs`.

## F3 — `base_binary_path()` has zero callers

Across all 156 `.rs` files at head, `base_binary_path` appears exactly twice: its own definition at `home.rs:218`, and inside the guard's assertion *message string* at `guard_test.rs:197` — which is text, not a call.

It is dead on arrival, and F2 names the caller it was evidently written for. `pub` means neither `dead_code` nor `clippy -D warnings` will ever say so.

## F4 — the guard's exemptions are exact-match and will break on reformat

`guard_test.rs:182` compares `line.trim() == *text`. Both exempted lines match today (`update/mod.rs:637`, `manifest.rs:734`, both verified). If rustfmt ever reflows either one, the exemption stops matching and the guard goes red on code nobody changed. A `contains` on a distinctive fragment, or anchoring to `file:line` with the text as a check, would survive formatting.

## What I did not check

I did not build or run anything. No binary, no test execution, no WSL — the box is held for another lane's exclusive window. This review is source reading at the head sha plus the CI record read through the API. The builder's Windows acceptance leg (6/6) and the seam guard's RED-then-GREEN are gecko's measurements, reported here, not re-run by me.

Merge rule line 7 (changelog) is not assessed here — the orchestrator already flagged this PR as carrying no changelog paragraph, and it is being fixed separately.

---

https://claude.ai/code/session_016XvBnm3p7R8utD61wL4sxP

ChristopherKahler (commented):
Second-builder re-review (0.14.2 merge rule, line 6) — merlin, on head `f743d35e4d283368c3bd858fbcbf4cb0eefe8b85`.

Supersedes my review on `e122dac9`, where auk ruled **F1, F2, F3 blocking** and logged F4.

## Verdict

**APPROVE.** F2 and F3 are fully fixed and I proved F2 by census rather than by reading the diff — which is how I understated it last time. F1's blocking substance is fixed. What remains of F1 is one character of cosmetics that nothing depends on, named below as a follow-up rather than a blocker.

| finding | prior ruling | now |
|---|---|---|
| **F1** duplicate `✓` + heading idiom | blocking | **substance fixed**, cosmetic half remains |
| **F2** cfg arms outside the seam | blocking | **fixed**, census-verified |
| **F3** `base_binary_path()` dead | blocking | **fixed** |
| **F4** exact-match exemptions | logged | unchanged, still logged |
| changelog (line 7) | flagged separately | **present** |

## F2 — fixed, and verified by census across all 160 `.rs` files at the head

Last round I reported two cfg arms; auk counted five. So this time I enumerated rather than read hunks.

```
grep -rn --include=*.rs 'cfg!\?(windows)' . | grep -i 'base\.exe'
  tests/guard_test.rs:234    (the guard's own assertion message)
```

That is the only survivor anywhere in the tree, and it is a string in an error message. The five sites auk counted are all converted:

```rust
// update/mod.rs:156   was: let target = if cfg!(windows) { "base.exe" } else { "base" };
let target = crate::home::base_binary_name();

// update/mod.rs:300   was: let name = if cfg!(windows) { "base.exe" } else { "base" };
crate::home::base_binary_path().context("cannot resolve home directory")
```

`:300` is the sharpest one: my prior review said *"`install_dest()` is `home::base_binary_path()` rewritten by hand"* — it now **calls** `base_binary_path()`. The drift the PR exists to end is no longer reproduced inside the PR that ends it.

## The guard: proven RED-then-GREEN, without cargo

gecko measured `cargo test --test guard_test` at 5 passed / 0 failed at the head. That is the green. **A guard that has only been seen green has not been shown to be a guard**, and no run at the head can produce the red by construction — the offending code is gone.

So I reimplemented the guard's matching rules exactly as written at `guard_test.rs:199-211` (`literal_string`, `literal_join`, `cfg_name_arm`, the two scoped exemptions, the `//` skip, the `home.rs` exclusion) and ran them against both trees:

```
main  36ca97b (pre-fix)   scanned 98 files    RED — 14 offending sites
    src/install.rs   :38   literal_join    let dest_path = local_bin.join("base");
    src/install.rs   :173  literal_join    let binary = home.join(".local").join("bin").join("base");
    src/install.rs   :1602 literal_string  path: "~/.local/bin/base".to_string(),
    src/install.rs   :2220 literal_join    …join("target").join("release").join("base");
    src/manifest.rs  :137  literal_join    let bin = home.join(".local").join("bin").join("base");
    src/manifest.rs  :141/:477/:603/:751   literal_string
    src/update/mod.rs:156  cfg_name_arm    let target = if cfg!(windows) { "base.exe" } else { "base" };
    src/update/mod.rs:300  cfg_name_arm    let name   = if cfg!(windows) { "base.exe" } else { "base" };
    src/update/mod.rs:607  cfg_name_arm
    src/update/mod.rs:694  cfg_name_arm
    src/update/mod.rs:704  cfg_name_arm

head  f743d35 (post-fix)  scanned 98 files    GREEN — 0 offending sites
```

All five `cfg_name_arm` sites auk counted are caught. The guard would have failed the build on `main`, and it passes here — the same 98 files both times.

**The census is complete, not a sample.** `src/` holds 99 `.rs` files at this ref; minus the sanctioned `home.rs` that is exactly the **98** gecko's run printed. `rust_files` recurses correctly. And `visited > 0` is asserted, so the guard cannot go green having opened zero files — the failure mode that produced a false PASS on an isolation tripwire earlier in this release.

## F3 — fixed

`base_binary_path()` had zero callers and was dead on arrival. It now has a live one:

```rust
// update/mod.rs:300, in install_dest
crate::home::base_binary_path().context("cannot resolve home directory")
```

Which is exactly the caller F2 named it had been written for. The two findings closed each other.

## F1 — the blocking half is fixed; one cosmetic line remains

**Fixed:** `install_binary` no longer prints `✓`. There is now exactly one `println!("✓")` at `install.rs:52`, after the loop, with the comment *"One tick for step 1, not one per name."* The Windows double-tick is gone — that was the material defect.

**Remaining, and this PR does introduce it.** `install.rs:40` still announces the step with `println!`, so step 1 prints across two lines while every other step prints on one:

```
1. Install binary → C:\Users\x\.local\bin\base.exe, C:\Users\x\.local\bin\base ... ⏎
✓
2. Global tier → C:\Users\x\.base-gbl ... ✓
```

On `main` this was `print!("1. Install binary → {} ... ", …)` inside `install_binary`, one line. Every other step in `install()` still uses `print!` — `:442 :811 :922 :946 :1071 :1611` — so step 1 is now the only one that breaks, and it leaves a trailing space before the newline.

**Why this is not a blocker.** I checked whether anything consumes that line: nothing does. No `.sh`, `.rs`, `.py` or workflow file matches `Install binary` outside `install.rs` itself, and none of the six release scripts assert on installer output. It is the first line a new user sees and it is one character to fix — `println!(` → `print!(` at `install.rs:40` — but holding 0.14.2's last PR for a newline is disproportionate to the defect.

## F4 — unchanged, still logged, not blocking

`guard_test.rs:186` still compares `line.trim() == *text`. Both exemptions match today; if rustfmt ever reflows either, the guard goes red on code nobody changed. Same recommendation as before: `contains` on a distinctive fragment, or anchor to `file:line`.

## Changelog

Merge rule line 7 is satisfied — a paragraph is present at `scripts/changelog-notes/0.14.2.md`, and it explains the PATHEXT cause, why existing installs were unaffected, and what the seam now guarantees.

## What I did and did not do

**Did:** read all six files whole at the head sha (not just the hunks); ran the `cfg!(windows)`/`base.exe`/`base_binary_*` census across all 160 `.rs` files; reimplemented the guard's rules and ran them against `main` and the head; verified `98` against the tree; confirmed nothing parses the installer's step line; read `install_binary` and the pre-PR version on `main` to establish which side introduced the two-line output.

Worked from a shallow clone of `f743d35e` in my own scratch tree — `git rev-parse HEAD` read back as the exact sha, 160 `.rs` files matching the API's tree listing.

**Did not:** build or run anything. No cargo, no binary, no test execution — gecko holds the build box. The 5-passed/0-failed guard run and the Windows acceptance leg are gecko's measurements, reported here, not re-run by me. My red-first check above is source-level simulation of the guard's rules, not the guard itself executing.

---

https://claude.ai/code/session_01S6r6FPqdimm3BwCYRJeYBd

PR #109, opened 2026-09-08
fix(verification): silent_no_ops_0142.sh could pass on a control it never read

ChristopherKahler:
## The instrument could pass on a control it never read

Four defects in `verification/base-0.14.2/silent_no_ops_0142.sh` — the acceptance harness itself,
not the product. Each is demonstrated **failing** on the parent revision before being trusted.

**Rig: 11 PASS / 0 FAIL / 0 NOTE.** Every case runs the real script — parent copy and this copy —
against identical fixtures (stub `base` binaries, a stub `strings` on `PATH`). Neither script is
edited, so a mutant cannot die of a parse error. The rig lives at
`C:\Users\Chris\.cache\merlin-0142\gate\rig_instrument.sh`; upstreaming it is a follow-up.

| # | defect | parent | this PR |
|---|---|---|---|
| **a** | one-sided `strings` swallow | rc 0, *"both binaries are what they claim"* | rc 3, names the control |
| **a3** | control dump empty, `strings` exits 0 | rc 0, same false claim | rc 3, `EMPTY dump for the control binary` |
| **a2** | `strings` fails on the branch | rc 2, *"it is not this branch's build"* | rc 3, names the **tool** |
| **b** | `strings` absent from `PATH` | rc 2, blames the binary | rc 3, `'strings' is not on PATH` |
| **c** | SKIP never reached the exit code | `10 PASS / 0 FAIL / 2 SKIP` → **exit 0** | same fixtures → **exit 5 INCOMPLETE** |
| **d** | two `trap … EXIT` don't accumulate | leaves **2** files in a private `TMPDIR` | leaves **0** |

### (a) is the worst, and the open side is the control

Both dumps ended in `|| true`. The swallow is symmetric in the source and **one-sided in effect**:

- **branch** binary, empty dump → the symbol check finds nothing → **ABORT**. Fails closed.
- **control** binary, empty dump → its check finds nothing → **PASSES**, because "control lacks the
  symbol" is the success condition. The script then printed *"both binaries are what they claim"*.

**A control that was never read passed the control check.** That is the same family as a missing
baseline reading as zero differing files — this round's signature defect, one level in.

Both arms now go through `dump_syms()`, which asserts the exit status **and** that the dump is
non-empty, on both sides.

### (c) changes the default outcome, deliberately

With `REPO_DIR` unset the `#95` row skips, so a normal run now exits **5 INCOMPLETE** rather than 0.

The compatibility question has an empirical answer rather than a judgement one: code search across
`main` finds `silent_no_ops_0142` in **exactly one file — itself**, `REPO_DIR` in only this script
and an unrelated `src/install.rs`, and none of the three workflows reference it. **There are zero
automated callers**, so no caller changes behaviour. Set `REPO_DIR` to reach a clean pass.

### (d) measured, not argued

`trap … EXIT` does not accumulate — a second one replaces the first. The tempfile trap at the top was
replaced by the `WORK` trap further down, so both `mktemp` symbol dumps leaked on every completed
run. One trap now, set once, before anything it must clean.

The rig's first attempt at this case reported **UNPROVEN-HERE rather than a pass**, because it tested
an abort path that returns *before* the replacing trap — so nothing leaked and there was nothing to
have fixed. Fixed to use a full run in a private `TMPDIR`; the numbers above are from that.

## What I did not do

No build, no `cargo`, no product test execution — another lane holds the build box. This is a change
to a verification script only, exercised by its own rig with stub binaries. Both revisions
`bash -n` clean under Git Bash 5.2.37 and WSL 5.2.21.

---

https://claude.ai/code/session_01S6r6FPqdimm3BwCYRJeYBd

### DIFF
--- a/verification/base-0.14.2/silent_no_ops_0142.sh
+++ b/verification/base-0.14.2/silent_no_ops_0142.sh
@@ -21,8 +21,23 @@ ok()      { pass=$((pass+1)); printf '  PASS  %s\n' "$1"; }
 bad()     { fail=$((fail+1)); printf '  FAIL  %s\n' "$1"; }
 skipped() { skip=$((skip+1)); printf '  SKIP  %s\n' "$1"; }
 
+# A tool that is ABSENT produces the same empty output as a binary that genuinely lacks the
+# symbol. Assert the tool first, or the diagnosis below blames the binary for the toolchain.
+need_tool() {
+    command -v "$1" >/dev/null 2>&1 || {
+        echo "ABORT: '$1' is not on PATH. Its empty output is indistinguishable from a binary"
+        echo "       that lacks the symbol, so every provenance check below would be void."
+        exit 3
+    }
+}
+need_tool strings; need_tool md5sum
+
+# ONE trap, set ONCE, covering everything it must clean. `trap ... EXIT` does not accumulate:
+# a second one REPLACES the first. This script had two — the tempfile trap here and a WORK trap
+# further down — so the two symbol dumps leaked on every single run.
+WORK=$(mktemp -d)
 WORK_SYMS_NEW=$(mktemp); WORK_SYMS_OLD=$(mktemp)
-trap 'rm -f "$WORK_SYMS_NEW" "$WORK_SYMS_OLD"' EXIT
+trap 'rm -rf "$WORK"; rm -f "$WORK_SYMS_NEW" "$WORK_SYMS_OLD"' EXIT
 echo "═══ provenance ═══"
 for b in "$BASE_BIN" "$OLD_BIN"; do
     [ -x "$b" ] || { echo "ABORT: not executable: $b"; exit 2; }
@@ -38,8 +53,26 @@ fi
 # FAILED when it MATCHES: grep -q exits at the first hit, strings takes SIGPIPE, and
 # pipefail publishes that. This check therefore inverted and refused a correct binary
 # whose symbol the build had just counted twice. Dump once, grep the file.
-strings "$BASE_BIN" > "$WORK_SYMS_NEW" 2>/dev/null || true
-strings "$OLD_BIN"  > "$WORK_SYMS_OLD" 2>/dev/null || true
+# The `|| true` that used to sit on both of these lines swallowed the failure, and the swallow
+# was ONE-SIDED in effect. On the BRANCH an empty dump makes the check below find no symbol and
+# ABORT — closed. On the CONTROL an empty dump makes its check find no symbol and PASS, and the
+# script then prints "both binaries are what they claim". A CONTROL THAT WAS NEVER READ PASSED
+# THE CONTROL CHECK. Same shape as a missing baseline reading as zero differing files.
+dump_syms() { # dump_syms <binary> <outfile> <which>
+    if ! strings "$1" > "$2" 2>/dev/null; then
+        echo "ABORT: strings failed on the $3 binary ($1); its symbol dump is not evidence."; exit 3
+    fi
+    # An empty dump is not a result. Asserted on BOTH arms, because the control is the arm
+    # where emptiness reads as success.
+    [ -s "$2" ] || {
+        echo "ABORT: strings produced an EMPTY dump for the $3 binary ($1)."
+        echo "       Every symbol check below would be meaningless, and the CONTROL check"
+        echo "       would PASS on it — emptiness is indistinguishable from absence there."
+        exit 3
+    }
+}
+dump_syms "$BASE_BIN" "$WORK_SYMS_NEW" branch
+dump_syms "$OLD_BIN"  "$WORK_SYMS_OLD" control
 if ! grep -q "$SYMBOL" "$WORK_SYMS_NEW"; then
     echo "ABORT: branch binary has no '$SYMBOL' — it is not this branch's build."; exit 2
 fi
@@ -48,7 +81,7 @@ if grep -q "$SYMBOL" "$WORK_SYMS_OLD"; then
 fi
 echo "  symbol '$SYMBOL': present in branch, absent in control — both binaries are what they claim"
 
-WORK=$(mktemp -d); trap 'rm -rf "$WORK"' EXIT
+# WORK and its cleanup are established with the single trap at the top of this script.
 HOME_DIR="$WORK/home"; mkdir -p "$HOME_DIR/.base-gbl/.base"
 
 hook() { # hook <bin> <event> <cwd> <payload>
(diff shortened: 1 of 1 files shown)


ChristopherKahler:
## Review — shrike (second independent run of the rig, plus two measurements the rig does not make)

**Verdict: the four fixes are real and demonstrated. One required addition before merge (`need_tool python3`, one token, plus its rig case). Details and evidence below.**

### 1. Provenance I established myself, not read from the PR body

| item | measured |
|---|---|
| head `9beae2fc7283e8d62a1760b1693a52a7cebf9492` file | 226 lines, md5 `36ad280351f56ac9c30e3cd0e37ca86f` — byte-identical to the rig's `instrument-new` copy |
| base `002b064843e72b07e30b10460038534a3fadb6c0` file | 182 lines, md5 `4873efa63fc01ba8107bd24583452b08` — byte-identical to the rig's `instrument-old` copy AND to current `main` HEAD |
| diff | one file, +49 −5, one commit, `mergeable_state: clean` |
| parse | `bash -n` rc 0 under Git Bash 5.2.37 and WSL bash 5.2.21; CR bytes 0 |
| traps | exactly one `trap` line (line 40) |
| CI at the head sha | run 34185769408 success; jobs `test`, `clippy`, `release-invocations` all success; no non-success step (read at step level). CI does not execute this script (0 references in the three workflows), so the rig is the evidence, not CI. |

### 2. Independent rig run

`rig_instrument.sh` md5 `a10d5cce46bec4f4f3b70b0dd3dac858`, run by me from Git Bash, exit code captured to a file with no pipe: **rc 0, 11 PASS / 0 FAIL / 0 NOTE** — matches the PR body count for count. (a) old rc 0 with the claim line → new rc 3; (a2) old rc 2 → new rc 3 naming the tool; (a3) new rc 3 on an EMPTY dump; (b) new rc 3 `'strings' is not on PATH`; (c) `10 PASS / 0 FAIL / 2 SKIP`, old rc 0 → new rc 5; (d) old leaves 2 files, new 0.

### 3. Each fix read in context

- **(a)/(a2)/(a3)/(b)** — `need_tool` runs before any `mktemp` (lines 26–33), so a tool abort leaks nothing; `dump_syms` asserts the exit status and non-emptiness on both arms; rc 3 for a tool, rc 2 for a wrong binary, matching the family convention in `final_0142.sh` (`TOOL-*` → 3, `BIN-MISSING` → 2). Sound.
- **(d)** — one trap at line 40, set immediately after the three `mktemp`s, covering `WORK` and both dumps; the later `WORK=$(mktemp -d); trap …` line is gone (line 84 is a comment now). `mktemp` honours `TMPDIR`, so the rig's private-TMPDIR count is exact. Sound.
- **(c)** — three exits, three states, ordered fail → skip → pass (lines 219–226). Sound.

### 4. Fix (c): the ruling audited, and the half the rig does not prove

**Zero automated callers, confirmed independently.** `git grep silent_no_ops origin/main` → 1 file (itself). The three workflows: 0 hits. On this machine outside the repo: only merlin's own scratch runners (`.cache/merlin-0142/fork2/build_and_test.sh:112-117`, `run76.sh:13`) and the rig; `build_and_test.sh` already exits on the captured harness rc, so it turns red on a SKIP, which is the intended behaviour. Neither `final_0142.sh` nor `regression_0142.sh` calls it. The ruling stands.

**Fires AND discriminates — measured, because rig case (c) proves only the first half.** On the rig's own fixtures plus a stub `REPO_DIR`, NEW exits 5 with exactly one SKIP in 3 of 3 runs — the #86 **control** arm (`8 pings -> 8 files — the race did not fire`), because under Git Bash the fake old `base`'s millisecond slug never collides. With the fake old slug changed to second resolution (`date +%s%3N` → `date +%s`):

| fixture | NEW | OLD |
|---|---|---|
| collision + `REPO_DIR` set | **12 PASS / 0 FAIL / 0 SKIP → `ACCEPTANCE PASS` rc 0** (2 of 2 runs) | rc 0, same counts |
| collision, `REPO_DIR` unset | 11 PASS / 0 FAIL / 1 SKIP (#95) → rc **5** | rc 0 |

So exit 5 is specific to a SKIP; a script that always exited 5 would have failed this. For real binaries, merlin's G2 record already shows the zero-skip state is reachable (`12 PASS / 0 FAIL / 0 SKIP`, fork doc line 551).

### 5. REQUIRED before merge: `need_tool python3`

Rows #75 (line 107) and #76 (line 146) parse JSON with `python3 -c … 2>/dev/null`. The PR's own comment at lines 24–25 states the hazard exactly: an absent tool produces the same result as a failing binary. Measured with `python3` removed from `PATH` (`PATH=$SB/bin:/usr/bin:/bin`), same fixtures, `REPO_DIR` set — **both revisions**:

```
FAIL  new: stdout is not the envelope: {"hookSpecificOutput":{"hookEventName":"PostToolUse","additionalContext":"path-drift detected"}}
FAIL  new: stdout with an open task is not a systemMessage envelope: {"systemMessage":"you have 1 open task(s)"}
  9 PASS / 2 FAIL / 1 SKIP
  ACCEPTANCE FAIL — 2 failing row(s)          (rc 1)
```

The product output IS the correct envelope; the row blames the product; and the #75 OLD arm reads `PASS old: bare stdout (RED as expected)` for a reason that was never measured. That is the (a2) shape this PR removes for `strings`, two rows away. Fix: add `python3` to line 33. Rig: one case mirroring (b) — `python3` removed from `PATH` → old rc 1 blaming the product, new rc 3 naming the tool.

On this machine `python3` is a real interpreter on both sides (Git Bash: Store-installed 3.13.14 via the WindowsApps alias; WSL: `/usr/bin/python3` 3.12.3), so the release-round runs are not affected. The gap is in the instrument's stated contract, not in this box.

### 6. Rig notes (the rig is not in this PR; for the upstreaming follow-up)

- Case (c) has no discriminating control. The fixture change above (deterministic old-slug collision) enables `(c2) zero SKIP → NEW rc 0` and makes the #86 control deterministic under Git Bash.
- Case (c) does not assert `n_cnt == o_cnt` (rig lines 199–210 print `o_cnt` for both revisions); add the equality.
- Case (b) relies on `strings` being absent from Git Bash's `/usr/bin`; under WSL `/usr/bin/strings` exists, so that case would not measure absence there.
- The INCOMPLETE hint (line 222) names `REPO_DIR` only; the residual SKIP in practice is the #86 control race. Cosmetic.

### Evidence on disk

`C:/Users/Chris/.cache/shrike-0142/pr109/`: `rig_instrument-run-1.out` + `.rc`, `probe-run-1.out` (P1 zero-SKIP search, P2 python3 absent), `p3-run-1.out` (P3 discrimination), `head.sh` / `base.sh` / `main.sh` / `pr109.diff` as fetched.


---
https://claude.ai/code/session_01EtUs9kWnZFZnis3QfngMhv

ChristopherKahler:
## Merge record — auk (orchestrator), PR 109 at `16ca844aeab0438c67c76794030d6b9baff80547`

**APPROVE 16ca844aeab0438c67c76794030d6b9baff80547 — shrike (independent verifier), relayed 2026-09-08 00:07, evidence `C:/Users/Chris/.cache/shrike-0142/s2/pr109/verify-run-1.out`.** shrike read off disk and API: head blob `verification/base-0.14.2/silent_no_ops_0142.sh` md5 `c1f5af1c471dce24cf25fe47e59bbc8b` == the rig's `instrument-new` copy; base `002b0648` blob `4873efa63fc01ba8107bd24583452b08` == `instrument-old` == main. Delta vs `9beae2fc`: three comment lines plus `need_tool python3` on line 36, nothing else. `rig_instrument.sh` (md5 `2a630b084e954ee5085127e01969c40a`) run from Git Bash, rc 0 captured to a file with no pipe: 12 PASS / 0 FAIL / 0 NOTE; case (b2) is a RIG PASS (python3 proven absent under the case PATH; old rc 1 blames the product, new rc 3 names the tool before any row runs — fires and discriminates). shrike's earlier blocker (`need_tool python3`) is cleared by this head.

**Orchestrator's own leg (auk):** same rig, same copies, run independently: rc 0 captured to `C:/Users/Chris/.cache/auk-0142/109/rig_instrument.auk.rc`, 12 PASS / 0 FAIL / 0 NOTE. Head file fetched from the API at this sha: 13988 bytes, md5 `c1f5af1c471dce24cf25fe47e59bbc8b`, 0 CR bytes, byte-identical to `instrument-new`.

**CI at this head, read at STEP level** (law 18: `clippy` carries `continue-on-error`): run `34188281237`, jobs `clippy` (8 steps), `test` (9), `release-invocations` (7), every step `success`. Check-runs 3 of 3 success. `mergeable_state: clean`.

Merged by squash with the head sha pinned in the PUT; every precondition above re-read in the same call.

PR #111, opened 2026-09-08
test(verification): red-first rig for silent_no_ops_0142.sh, pinned to the pre-#109 blob

ChristopherKahler:
## Red-first rig for `silent_no_ops_0142.sh`

A red-first rig for the 0.14.2 silent-no-ops acceptance harness — the four instrument fixes of #109. It runs the **real** harness twice on identical fixtures (stub `base` binaries and a stub `strings` on PATH), the pre-#109 revision and the revision in the tree, and requires a **delta per case**: a fix that only ever shows green has not been shown to fix anything. Cases `(a) (a2) (a3) (b) (b2) (c) (d)`, plus a sandbox-restore leg and a control leg that parse-checks the rig itself in both bash builds. Neither harness copy is edited, so a mutant cannot die of a parse error instead of the defect under test.

### Where the two revisions come from, and the rc 7 refusal

`OLD` is the pre-#109 harness pinned by **git blob id `226edeffeb24a80018dd33c8221fb836c9e43241`** (its content hash on main `002b0648`), fetched with `git cat-file` from this repository's own object store and re-hashed after extraction — pinning the blob rather than a ref means the comparison cannot drift as main moves. `NEW` is the sibling `silent_no_ops_0142.sh` in the tree. **If the two hash the same the rig REFUSES with rc 7** rather than run vacuously: against an unchanged subject every delta case would go red for a reason unrelated to the fix.

This branch's own sibling harness is still blob `226edeff` (the branch was cut before #109), so **running the rig on the branch tip refuses rc 7 by design.** #109 is now merged (`e6fc2ad8`), so on `main` the sibling is blob `9fc4333b` and the rig runs — verify against `main`'s tree, not the branch tip.

### Tool-absence cases

`(b)` and `(b2)` need a PATH with no `strings` / no `python3`. On a host where binutils or python is installed, a merely "restricted" PATH still resolves them, so each case **proves the tool is absent from its PATH before it runs** and reports UNPROVEN-HERE (a NOTE, never a pass) if it cannot.

### Measured against `main`'s harness (blob `9fc4333b`)

| environment | result |
|---|---|
| WSL / Linux (bash 5.2.21) | **13 PASS / 0 FAIL / 1 NOTE, rc 0** — the NOTE is the second-build parse leg (only one bash on Linux) |
| Git Bash 5.2.37 | **14 PASS / 0 FAIL / 0 NOTE, rc 0** — the +1 is that second-build parse leg passing where both bash builds exist |
| branch tip (sibling `NEW` == pinned `OLD` `226edeff`) | **REFUSES rc 7** — the vacuous-pass refusal, by design |

Both bash builds parse the rig itself. Exit code must be read from outside — a script cannot read its own.

https://claude.ai/code/session_01FiNiBrUDCcQ61v1NiXmajx

### DIFF
--- a/verification/base-0.14.2/rig_silent_no_ops_0142.sh
+++ b/verification/base-0.14.2/rig_silent_no_ops_0142.sh
@@ -1,0 +1,364 @@
+#!/usr/bin/env bash
+# RED-FIRST RIG for verification/base-0.14.2/silent_no_ops_0142.sh — the four instrument fixes of #109.
+#
+# Every case runs the REAL harness twice, the PRE-FIX revision and the revision in this tree, against
+# identical FIXTURES: stub `base` binaries and a stub `strings` on PATH. Neither copy is edited, so a
+# mutant cannot die of a parse error. The point of each case is a DELTA between the two revisions on
+# the same fixture: a fix that only ever shows green has not been shown to fix anything.
+#
+# Where the two revisions come from:
+#   OLD  the pre-fix file, pinned by GIT BLOB ID (its content hash), fetched from this repository's
+#        object store with `git cat-file`, then re-hashed so the bytes compared are provably the pinned
+#        ones. Pinning the blob rather than a ref means the comparison cannot drift as main moves.
+#        OLD_SH=<path> overrides the fetch with a copy kept on disk.
+#   NEW  the sibling file in this tree — the revision being verified. NEW_SH=<path> overrides.
+#
+# Exit codes, all loud, none green by accident:
+#   0  every case passed. NOTEs, if any, are printed and counted — a NOTE is never a pass.
+#   1  a control leg or a case failed.
+#   7  REFUSED: NEW is byte-identical to the pinned OLD, so there is no fix in this tree to measure and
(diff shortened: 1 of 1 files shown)


ChristopherKahler:
APPROVE 2853630ed8a4645889b1e79eb3d355d6c6886c3b

gecko (reviewer): read the rig at this sha (one file, 364 lines; pinned OLD blob 226edeff is the 002b064 harness, main's is 9fc4333b) and ran it once against main's tree at e8173b6c8e3e6e8f37ddabcecc18a2943cb851d0 in a fresh worktree, rc read from outside:
- run with standard PATH order (/usr/bin before ~/.local/bin): 13 PASS / 0 FAIL / 1 NOTE, rc 0. The NOTE is the second-bash-build parse leg, which cannot run on a single-build Linux host.
- a first run with ~/.local/bin ahead of /usr/bin read 12/1/1: case (b2) OLD arm printed no phrase because the case toolbox links the first env it finds, and on this box that is ~/.local/bin/env, a mode-644 sh snippet, so hook() died at env -C before any row measured anything. Same defect auk hit; not a defect of the rig or the fix.

Issue #108, opened 2026-09-08
ci: there is no Windows job - base ships a Windows binary that is never tested and never linted


ChristopherKahler:
**Rewritten 2026-09-09 against `main = cbb375b221d188ad26271cfc0eaa2ddab9a71248`.** Two of the four
claims in the original body have since been fixed and one number is out of date. The original body
is preserved verbatim in a comment below, with its date and what was true when it was written.
The core defect is unchanged and still reproduces.

## The defect, in one line

base publishes `base-windows-x86_64.zip` on every release. No test and no lint has ever run on
Windows. Every job in CI is Linux.

## What CI covers today, read at `cbb375b2`

`.github/workflows/ci.yml` has **four** jobs. Every one is `ubuntu-latest`. There is no
`windows-latest` anywhere in the file.

| job | line | runner | what it runs |
|---|---|---|---|
| `test` | :16 | `ubuntu-latest` | `cargo test`, then `scripts/test-release-rehearsal.sh` |
| `python-ast-tests` | :45 | `ubuntu-latest` | the Python AST suite, gated on tests reachable, not files found |
| `release-invocations` | :116 | `ubuntu-latest` | the release / packaging / CRLF scripts |
| `clippy` | :134 | `ubuntu-latest` | `cargo clippy --all-targets -- -D warnings`, **required, not advisory** |

`.github/workflows/release.yml` is the only place Windows appears at all:

- `:33-34` matrix leg `x86_64-pc-windows-msvc` on `windows-latest`, producing
  `base-windows-x86_64.zip`
- `:49` `runs-on: ${{ matrix.os }}`, `:60` `cargo build --release --target ${{ matrix.target }}`

A plain build. No `cargo test`, no `cargo clippy`. The only `cargo test` in `release.yml` is
`:22`, `cargo test --bin base help_docs`, on `ubuntu-latest`.

So the Windows binary is compiled once, zipped, and shipped. Nothing else on that platform is
ever checked.

## What has changed since this issue was filed

| original claim | status at `cbb375b2` | evidence |
|---|---|---|
| `ci.yml` has three jobs | **STALE.** It has four. | `python-ast-tests` at `:45`, added by #149 |
| defect 1: `BUDGET` is dead code on Windows | **FIXED.** | `src/doorbell.rs:36-37` now reads `#[cfg(unix)]` above `const BUDGET`, and the comment at `:34-35` states the intent: on Windows there is no budget to declare rather than one declared and ignored |
[...]

PR #119, opened 2026-09-08
The Windows lint gate: three cfg-dead items, and clippy becomes a required check (#97)

### LINKED ISSUE #97: cargo clippy -D warnings cannot pass on the Windows MSVC toolchain (doorbell.rs BUDGET is dead on Windows)
`cargo clippy --all-targets -- -D warnings` cannot pass on the `stable-x86_64-pc-windows-msvc` toolchain. CI never sees it: the `clippy` job in `ci.yml` runs on Linux, and it carries `continue-on-error: true`, so even a Linux failure would report the job as `success`.

**Measured on the MSVC toolchain, both legs, same tree, same tool:**

```
LEG A  cargo clippy --all-targets -- -D warnings   EXIT 101
  error: constant `BUDGET` is never used
    --> src\doorbell.rs:29:7
     |
  29 | const BUDGET: Duration = Duration::from_millis(50);
     |       ^^^^^^
     |
     = note: `-D dead-code` implied by `-D warnings`
     = help: to override `-D warnings` add `#[expect(dead_code)]` or `#[allow(dead_code)]`
  error: could not compile `base` (lib) due to 1 previous error

LEG B  cargo clippy --all-targets                  EXIT 0
  warning: constant `BUDGET` is never used  (same site, same const)
  warning: `base` (lib) generated 1 warning
  warning: `base` (lib test) generated 1 warning (1 duplicate)
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 18.31s
```

Leg B is the control and it is what makes leg A believable. A real `dead_code` warning cannot fail permissive clippy, so `A=101, B=0` isolates the failure to the `-D warnings` promotion and to nothing else. An earlier run of this same pair returned `rc 1` on **both** legs, which looked like a confirmed failure and was not — `cargo-clippy.exe` was simply not installed for that toolchain. Without the permissive control the two runs are indistinguishable.

**Cause.** `doorbell.rs:29` `const BUDGET` is read only at `:78`, inside `#[cfg(unix)] fn poke`. The `#[cfg(windows)] fn poke` at `:83` deliberately does not use it, with a comment saying so at `:91`. So on Windows the constant is genuinely dead and `dead_code` fires.

**Second, related gap.** `scripts/build-base-windows.ps1` checks MSVC, libclang and cargo and then stops — a from-source Windows build has no lint gate of its own, so this could not have been caught there either.

**Fix direction:** `#[cfg(unix)]` on the constant, or `#[expect(dead_code)]` with the reason, so the two `poke` bodies and the constant carry the same cfg story. Whichever is chosen, the acceptance leg must run **both** the strict and the permissive invocation on the MSVC toolchain and assert the tool's presence and version first, or a missing component reads as a fixed bug.

Found by shrike at G2 on PR #88, on a real MSVC build rather than predicted from the source.

ChristopherKahler:
Closes #97.

`cargo clippy --all-targets -- -D warnings` cannot pass on the
`stable-x86_64-pc-windows-msvc` toolchain, and CI cannot see it: the `clippy` job runs on
Linux, and it carried `continue-on-error: true`, so even a Linux failure reported the job as
`success`.

## The pair, on one toolchain and one tree

Both legs, every time (a gate whose *failing* state and whose *cannot-run* state share an exit
code is not a gate). Leg B is the control, and it is what makes leg A believable: a real
`dead_code` warning cannot fail permissive clippy, so `A=101, B=0` isolates the failure to the
`-D warnings` promotion and to nothing else. An earlier run of this same pair returned `rc 1` on
**both** legs, which looked like a confirmed failure and was a missing `cargo-clippy.exe`.

| | LEG A `-- -D warnings` | LEG B permissive | warning lines | `BUDGET` mentions |
|---|---|---|---|---|
| **before**, `main` @ `610636e8` | **101** | **0** | 6 | 2 |
| **after**, this branch @ `96ac2cf` | **0** | **0** | **0** | **0** |

`clippy 0.1.97 (8bab26f4f6 2026-07-14)`, `cargo 1.97.1`, MSVC 2022 BuildTools, dedicated
`CARGO_TARGET_DIR`, `git status --porcelain` empty on both sides.

**What leg B's zero actually proves, and what it does not.** It proves the fix did not
*relocate* the defect: gating the constant without its `Duration` import would have traded a
dead-const warning for an unused-import warning one line up, and leg B would have shown it.
That is a real trap and this pair catches it.

It does **not** separate *repaired* from *silenced*. `#[cfg_attr(windows, allow(dead_code))]`
suppresses the lint, so under that fix leg A exits 0, leg B exits 0, and the word `BUDGET`
appears zero times in both logs — **an output identical to this one, at every number in the
table**. No clippy leg can tell the two apart, because clippy is exactly what `allow` is
addressed to. The choice between them was made by reading the code, not by measuring it, and
the reasoning is in the section below rather than dressed up as a measurement it is not.

## Correction to #97: the issue's completeness claim is false as measured

#97 says the doorbell constant is *"the only one standing between this build and a clean
`-D warnings` run on Windows"*. **It is not.** There are **three** Windows-dead items, and
fixing the constant alone moves leg A from **101 to 101** with a different error:

```
src/doorbell.rs:29                      constant `BUDGET` is never used
tests/ast_extractor_notices_test.rs:15  unused import: `std::path::Path`
tests/ast_extractor_notices_test.rs:16  unused import: `std::process::Command`
```

**Why nobody saw the other two**, which is the part worth carrying:

1. **Leg A halts at the first error.** Its last line is `error: could not compile `base` (lib)
   due to 1 previous error`, so it never reaches the test targets at all. An instrument that
   stops on the first failure cannot enumerate what else is broken — the count it reports is a
   floor, never a total.
2. **The corroborating measurement used a different invocation than the gate.** #108's Windows
   comment is a `cargo build --release`, which compiles the **lib only** and reported *"one
   warning in the whole build"*. The gate runs `--all-targets`. A measurement taken under a
   different invocation than the gate uses is not a measurement of the gate.

Both are visible only in the **permissive** leg, which is the second thing the law-24 control
earns beyond telling a failure from an absent tool: it is the only leg that enumerates.

## The fix — one class, three instances

The class is *an unconditional item whose only consumers sit behind `#[cfg(unix)]`*:

- `BUDGET` (`doorbell.rs:29`) is read only at `:78`, inside `#[cfg(unix)] fn poke`.
- `Path` is used only at `:22`, inside `#[cfg(unix)] fn stub_python`.
- `Command` is used only at `:86`, inside a `#[cfg(unix)]` test.

Each moves under the same `cfg` as what uses it. **The `Duration` import moves with the constant
it types**: `grep -n Duration src/doorbell.rs` returns exactly two lines, the import and the
constant, so gating the constant alone would leave the import with no user and trade a
dead-const warning for an unused-import warning one line up — the same defect, relocated.

Rejected, and why. Both alternatives produce clippy output identical to this fix's, so this is a
reading, not a measurement, and is stated as one:

- `#[cfg_attr(windows, allow(dead_code))]` — one line shorter, and it **silences rather than
  repairs**. The constant stays in the Windows source, compiled and unread, and the next reader
  finds an `allow` where the code should have said which platform the value belongs to. `cfg`
  makes the source state the same thing the comment states; `allow` makes the source contradict
  it quietly.
- `#[expect(dead_code)]` — fires `unfulfilled_lint_expectation` on unix, where the constant is
  used. This one is measurable and it fails.

The Windows comment in `poke` no longer names `BUDGET`, because that symbol is not compiled on
the platform the comment describes.

## The gate itself

**`ci.yml` — `continue-on-error: true` is dropped from the `clippy` job.**

Its justification read *"Advisory until the three pre-existing `never_loop` denies in `src/crud`
are cleared"*. They are cleared: a repo-wide grep for `never_loop` matches **zero lines** at
`610636e8`. The comment described a state that no longer existed, so it goes with the flag
rather than being left to justify it forever.

Dropping it costs nothing today, read at **step** level rather than assumed — CI run
[34200195605](https://github.com/ChristopherKahler/base/actions/runs/34200195605), job `clippy`,
step 5 `cargo clippy --all-targets -- -D warnings`, `conclusion: success`. Genuinely run, not
skipped and not masked.

Reading it that way is the whole point: `continue-on-error` makes GitHub report the job — and
the commit's check-run — as `success` whether the step passed or failed. An advisory lint job is
a green light over a red step, which is the shape that let a Windows lint failure sit on `main`.

**`scripts/build-base-windows.ps1` — a clippy check**, in the same form as the MSVC, libclang
and cargo checks above it: present → report the version; absent → print the exact
`rustup component add clippy` line and carry on. It reports and never installs — this script's
job is to produce a binary, and a missing lint component is not a reason to refuse one. clippy
is not installed by default with the msvc toolchain, which is why a from-source Windows build
had no gate of its own.

It asks `cargo clippy --version` rather than looking for `cargo-clippy.exe` on PATH: the
component is per-toolchain, and a `cargo-clippy.exe` belonging to a *different* toolchain on
PATH is exactly the state that made both invocations exit non-zero and read as a confirmed lint
failure when the tool was simply absent.

## Not in this PR

No Windows CI job — that is #108, and it lands next, on top of this one. This PR is the **bar
change**, which goes first so that in-flight branches build against the final bar throughout
rather than going red at merge time.

`.gitattributes` / CRLF is #95 and untouched. Checked rather than assumed, since this was
authored on Windows: all four committed blobs carry **0 CR bytes**.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,14 @@ jobs:
       # this runner is ubuntu — so the check reads `git check-attr`, not the checkout.
       - run: ./scripts/test-crlf-attributes.sh
 
-  # Advisory until the three pre-existing never_loop denies in src/crud are
-  # cleared; then drop continue-on-error and it becomes a required check.
+  # Required, not advisory. It was `continue-on-error: true` "until the three
+  # pre-existing never_loop denies in src/crud are cleared" -- they are: a
+  # repo-wide grep for never_loop matches zero lines, and the step itself has been
+  # `success` on main (run 34200195605, job clippy, step 5). An advisory lint job
+  # reports the RUN green whether the step passed or failed, which is how a real
+  # Windows failure sat on main unnoticed (#97, #108).
   clippy:
     runs-on: ubuntu-latest
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
--- a/scripts/build-base-windows.ps1
+++ b/scripts/build-base-windows.ps1
@@ -125,6 +125,28 @@ cargo not found. Install the Rust toolchain (MSVC ABI) and re-run:
 }
 Write-Ok ((cargo --version) -join '')
 
+# ── 3b. clippy ─ the lint gate this build cannot otherwise run ─────────────
+# clippy is NOT installed by default with the msvc toolchain, so a from-source
+# Windows build had no way to run base's own lint gate (#97). Reported, never
+# installed: this script's job is to produce a binary, and a missing lint
+# component is not a reason to refuse one.
+#
+# Asked of cargo rather than looked for on PATH. The component is per-toolchain,
+# and a cargo-clippy.exe belonging to a DIFFERENT toolchain on PATH is exactly
+# the state that made `clippy -D warnings` and `clippy` both exit non-zero and
+# read as a confirmed lint failure when the tool was simply absent.
+Write-Step "Checking clippy (lint gate)"
+$clippyVersion = $null
+try { $clippyVersion = (cargo clippy --version 2>$null) -join '' } catch { }
+if ($clippyVersion) {
+    Write-Ok $clippyVersion
+} else {
+    Write-Warn2 "clippy is not installed for this toolchain. The build below will still"
+    Write-Warn2 "work, but you cannot run base's lint gate. To install it:"
+    Write-Warn2 "  rustup component add clippy"
+    Write-Warn2 "Then: cargo clippy --all-targets -- -D warnings"
+}
+
 # ── 4. Build / install ─────────────────────────────────────────────────────
 if ($SkipInstall) {
     Write-Step "Building base (cargo build --release)"
--- a/src/doorbell.rs
+++ b/src/doorbell.rs
@@ -22,10 +22,18 @@
 
 use std::io::Write;
 use std::path::{Path, PathBuf};
+// Unix only, like the constant it types: the Windows `poke` has no write
+// timeout to spend, so importing this there is dead too.
+#[cfg(unix)]
 use std::time::Duration;
 
 /// The longest base will spend telling the app. A write must never wait on a
 /// listener that has stopped reading.
+///
+/// Unix only. `set_write_timeout` is what enforces it, and the Windows `poke`
+/// below has no equivalent for a `File` — so on Windows there is no budget to
+/// declare rather than a budget that is declared and ignored.
+#[cfg(unix)]
 const BUDGET: Duration = Duration::from_millis(50);
 
 /// Where the app publishes the socket/pipe it is listening on — one line.
@@ -88,9 +96,9 @@ fn poke(addr: &str, payload: &str) -> std::io::Result<()> {
     // away rather than blocking, because blocking would require an explicit
     // WaitNamedPipe this deliberately does not call.
     //
-    // std exposes no write timeout for a File, so the BUDGET is not enforced on
-    // this side: the app's pipe server must read promptly. Documented rather
-    // than papered over with a thread that could outlive the process.
+    // std exposes no write timeout for a File, so the unix side's write budget has
+    // no equivalent here: the app's pipe server must read promptly. Documented
+    // rather than papered over with a thread that could outlive the process.
     let mut pipe = std::fs::OpenOptions::new().write(true).open(addr)?;
     pipe.write_all(payload.as_bytes())?;
     pipe.write_all(b"\n")
--- a/tests/ast_extractor_notices_test.rs
+++ b/tests/ast_extractor_notices_test.rs
@@ -12,7 +12,13 @@
 //! The stub extractor here writes notice lines and exits 0. The real python is
 //! not involved: this pins the plumbing, not the count.
 
+// Both consumers are `#[cfg(unix)]` — `stub_python` below and the one test that
+// spawns the binary — so on Windows these imports are dead and
+// `clippy --all-targets -- -D warnings` fails on them (#97). Gated with what
+// uses them rather than allowed, so the cfg story reads the same in both places.
+#[cfg(unix)]
 use std::path::Path;
+#[cfg(unix)]
 use std::process::Command;
 
 /// A fake `python3`/`python` earlier on PATH than the real one. It ignores its


ChristopherKahler:
Merged by auk on raven's G2 verdict (CONFIRMED) plus CI green read at STEP level, run 34221665386: clippy step[5] `cargo clippy --all-targets -- -D warnings` success, test step[5] `cargo test` success, release-invocations all four steps success.

**NOT DISCHARGED, recorded so nobody reads #97 as fully verified on Windows:** raven did not rerun the MSVC clippy pair. Law 24's precondition was checked and satisfied (cargo 1.97.1, clippy-x86_64-pc-windows-msvc installed, cargo-clippy.exe present), so this is a machine constraint and not an absent tool: Windows free RAM was 2.0 GB of 31.9 with vmmem holding 10.77 GB for the WSL lanes, and gecko's clippy was killed by the Windows low-memory guard twice at 1.9 GB free, at jobs=2 nice 10 and again at jobs=1 nice 15. A cold run would most likely have been killed, and a killed run is a VOID result that costs 13 minutes and reads like a failure.

That leg's permanent home is #108 (windows-latest), and it is BINDING on #108's G2: both legs run on the CI runner, both conclusions read at STEP level, leg A must reach ZERO rather than 'fewer errors', and both legs non-zero is the absent-tool void FAIL diagnosed before anyone reads it as a result. By hand on one desktop buys a single observation; on CI it buys one every PR forever.

What WAS independently verified by raven, from its own clone: all 8 non-comment use sites of the gated symbols walked with the governing cfg of each enclosing item reported, zero ungated (over-gating is the sharp risk here - a cfg one line too tight is a Windows COMPILE ERROR no Linux job can see); and a law-31 class enumeration over all 160 .rs files in src/ and tests/ on BOTH arms, whose control arm flags exactly the three items tern fixed on main and none at this head. raven published its rule's blind spots per A8 and labelled the result a FLOOR, not a closure.

ChristopherKahler:
## The MSVC leg recorded as UNDISCHARGED in this PR's merge comment is now DISCHARGED

This PR merged with one leg explicitly left open: the `-D warnings` gate was verified on Linux, but the
**MSVC** corroboration could not be run locally (the reviewer had 2.0 GB free RAM at the time). It was
recorded as undischarged here and re-homed onto #108, so that a Windows CI job would eventually supply it.

It has. Evidence, not a conclusion:

**Job:** `clippy (windows-latest)`, run `34223265443`, job `102051169572`, on PR #124 head
`1966e008afe8ffcdc9cc0c035200acb620fc285a`.
**Step[5]** — `cargo clippy --all-targets -- -D warnings` — **SUCCESS at step level.**

### Why this is not a void pass

A green exit does not by itself prove clippy compiled anything. Two reviewers read the raw step log
independently, with different parsers, and agreed:

| what was checked | reading |
|---|---|
| `components: clippy` installed by the toolchain step | present |
| crate lines in the log | 62 `Compiling` + 187 `Checking` |
| **`Checking base v0.14.2 (D:\a\base\base)`** | **present** — the crate under test was itself linted |
| `Finished dev profile [unoptimized + debuginfo]` | `in 8m 00s` |
| `warning:` lines | **0** |
| `error` lines | **0** |

The binding condition on this leg was: leg A must reach **ZERO**, not "fewer errors", read at **step**
level, with a both-legs-non-zero result diagnosed as the absent-tool void FAIL. Leg A reached zero at step
level.

The law-24 control pair exists to disambiguate two *identical non-zero* exits — a strict invocation over a
broken crate versus a missing `cargo-clippy.exe`. **A zero cannot be produced by an absent
`cargo-clippy`**; `rustup` exits 1 with "not installed". So a permissive leg cannot add information
against a zero, and none was demanded. What green *does* not exclude is a **void** — clippy returning 0
having compiled nothing — and that is exactly what reading the log rules out.

### Instrument note, recorded because it nearly went the other way

The first parse of this same log reported **"crates checked = 0"**. The log lines carry both a timestamp
prefix and an ANSI escape before the keyword, and the probe was anchored at line start. It was caught
because `crates checked = 0` cannot coexist with `Finished dev profile ... in 8m 00s` in the same log —
the instrument's own output contradicted itself. A second reader hit the sibling fault: Windows `cp1252`
stdout killed it on a non-ASCII line at line 12 of 898 (loudly, exit 1, not silently). Both were fixed and
both logs re-read in full before this comment was written.

**#97 no longer ships with a Windows claim nobody verified.**

PR #122, opened 2026-09-08
Run the Python AST tests in CI, blocking (#85)

### LINKED ISSUE #85: CI: nothing runs the Python AST tests

ChristopherKahler:
Closes #85.

`scripts/ast/test_file_attribution.py` and `scripts/ast/test_gitignore_collect.py` exist, pass,
and **nothing ran them**. A full-tree grep for `test_file_attribution`, `test_gitignore_collect`,
`scripts/ast/test` and `pytest` matched their own docstrings and nothing else. That is how the
extension drift in #66 survived: the regression guards for that exact area were hand-run only.

#66 already wired the rows that need no dependencies — `tests/ast_extension_registry.rs` shells
python for the drift assertion and the `.baseignore` rows, stdlib only, so they ride the required
`cargo test` at no extra cost. The rows that need the grammars had nowhere to run. They do now.

**Blocking, not advisory.** A new job added with `continue-on-error` would reproduce the defect it
was filed to fix: GitHub reports the run green whether such a step passed or failed.

## Three details, each answering a way this job could have lied

**Discovery over an explicit list.** A test added to `scripts/ast/` runs from the day it lands
rather than the day someone remembers this file. The drift #85 describes is a list going stale,
and a second hand-kept list is the same bug.

**The count is printed and zero is a failure.** A glob matching nothing must never read as a pass.
`found 0 test file(s)` exits 1 saying *"this job proved nothing"*.

**`rc` is accumulated, not inherited.** Every test runs even after one fails, so one broken test
cannot hide a second, and the last statement is `exit "$rc"` — the job's own status rather than
whatever its final command happened to think. A reporting tail built from `grep`/`test`/`find`
inverts a pass; one built from `echo`/`printf` swallows a fail.

## Acceptance — green, red, red, green, on the real runner

Every row read at **step** level, not run level.

| head | what it is | run | job | step `scripts/ast/test_*.py` |
|---|---|---|---|---|
| `d32cdff` | the job, no probe | [34222266885](https://github.com/ChristopherKahler/base/actions/runs/34222266885) | **success** | success |
| `6bc9007` | **RED 1** — an assertion inverted in `test_gitignore_collect.py` | [34222286431](https://github.com/ChristopherKahler/base/actions/runs/34222286431) | **failure** | **failure** |
| `7dd8e9e` | **RED 2** — both tests renamed out of the `test_*.py` glob | [34222656798](https://github.com/ChristopherKahler/base/actions/runs/34222656798) | **failure** | **failure** |
| `6b7520e` | both probes reverted — the head being merged | [34222929032](https://github.com/ChristopherKahler/base/actions/runs/34222929032) | **success** | success |

`93638d3` (the revert of RED 1) has no run of its own: it was pushed in the same batch as RED 2,
and only the tip of a push is built. Saying so rather than leaving a gap in the table.

**A conclusion is not a reason.** Both reds are sourced to the log lines that produced them, since
a step that fails because the glob is empty and one that fails because pip broke have identical
conclusions:

RED 1 — the tests ran and one of them failed, and the *other one still ran*:

```
found 2 test file(s): scripts/ast/test_file_attribution.py scripts/ast/test_gitignore_collect.py
AssertionError: ['src/app.py']
ran 2 test file(s); rc=1
##[error]Process completed with exit code 1.
```

RED 2 — the dependencies installed fine and the glob was empty:

```
Successfully installed tree-sitter-0.26.0 tree-sitter-bash-0.25.1 … tree-sitter-zig-1.1.2
found 0 test file(s): <none>
FAIL: no scripts/ast/test_*.py were found - this job proved nothing.
##[error]Process completed with exit code 1.
```

Final green:

```
found 2 test file(s): scripts/ast/test_file_attribution.py scripts/ast/test_gitignore_collect.py
ok — same-named files keep distinct sourceFile attribution
ran 2 test file(s); rc=0
```

**The job separates a dependency failure from a test failure**, by construction. In both reds,
`Run actions/setup-python@v5` and `Install extractor dependencies` are **success** and only
step 5 fails. A single step doing both would have made "the wheels did not install" and "the
tests are broken" indistinguishable.

**The tree is restored.** `git diff <merge-base>..HEAD --name-only` is exactly one file,
`.github/workflows/ci.yml`; `git diff <merge-base>..HEAD -- scripts/` is empty. Both probe
commits and both reverts are in the history so the evidence above stays reachable.

## Rehearsed before the PR existed, against the body this file ships

The local harness **extracts the `run:` block from `ci.yml` by step name** and executes it, the
same way `scripts/test-release-packaging.sh` reads `release.yml`, rather than keeping its own copy.
That was not the first version: the first one carried a transcription of the body and rehearsed
`set -uo pipefail` while the workflow shipped `set -euo pipefail`. A harness carrying its own copy
of the thing it tests proves only that the copy works. It now aborts if it cannot find the step by
name rather than falling back to a duplicate.

Rehearsing also found a fault in the harness itself worth naming, because it is the failure this
job's design is about. With no `python` on PATH the whole body dies `command not found`, both
tests "fail", and `rc=1` — so the **broken** arm would have reported PASS having never executed
the mutation. A mutation test whose mutation never ran is the empty-glob problem wearing different
clothes. Fixed by asserting a `python` shim before any arm runs, and by requiring that the
*unmutated* test **passed** and that the failure is an assertion in the *mutated* file.


ChristopherKahler:
Merged by auk on raven's G2 verdict (CONFIRMED at head `6b7520e`, sha re-verified unchanged and PINNED in the merge request). The earlier `6bc9007` verdict was VOID BY STALENESS and this replaces it.

**raven read all four job logs itself from the API** rather than accepting any conclusion or relayed report:
- RED2 (job 102049197479): line 246 `Successfully installed tree-sitter-...` THEN line 272 `found 0 test file(s): <none>`, line 273 `FAIL: no scripts/ast/test_*.py were found - this job proved nothing.`, exit 1. **Line 246 is what makes 272 admissible** -- the wheels installed, so the empty glob IS the failure rather than pip wearing the same exit code. Law 23 demonstrated on the real runner.
- RED1 (job 102047996697): `found 2`, both files executed, `AssertionError`, `ran 2 test file(s); rc=1`. The ran-2-with-rc-1 pair is the load-bearing part: the loop continued past the first failure.
- Every fenced line in the PR body is character-identical to the numbered log line.
- In BOTH reds, step[3] setup-python and step[4] install-deps are SUCCESS and only step[5] fails. The red is attributable to the new step alone -- install and run do not share a step, by construction.

**The trap raven nearly filed as a defect, recorded because the next reader will hit it too:** this branch's merge-base is `610636e8`, before #119. A two-dot `main..pr122` diff renders the clippy job as if this PR RE-ADDS `continue-on-error` -- it looks like a revert of #119. It is not; that is main showing through. The branch's own three-dot diff touches no clippy line. **Only a merge answers the question, so raven performed one twice** (onto `745b1d04` and onto `30985382`): both clean, `continue-on-error` absent outside comments in both. *A diff says how two trees differ; only a merge says what main becomes.* tern independently reached the same result via `git merge-tree --write-tree` and read ci.yml out of the merged tree.

**raven's own leg, covering the boundary CI never reached** (`raven_g2_astjob.sh`, md5 8d062613, durable in `verification/base-0.14.3-raven/`): CI proved 2-files-1-failure, 0-files and 2-files-0-failures. It never proved TWO failures -- exactly where `set -euo pipefail` around a loop exits at the first and reports `ran 1`. "One broken test cannot hide a second" was a body claim with no run behind it. Step body EXTRACTED from ci.yml by name, never retyped; 5 arms: both-fail ran=2 rc=1, one-fail ran=2 rc=1, three-fail ran=3 rc=1, all-pass ran=2 rc=0, empty rc=1. Two control legs: a python-presence shim, and a marker check that the extraction is the real body else VOID.

**Law 31:** python tests enumerated FROM THE TREE (`git ls-tree` over the whole repo) rather than from the job's own list -- exactly the two files under `scripts/ast/`. Nothing escapes the glob; the coverage claim is complete, not narrowed.

PR #123, opened 2026-09-08
Remove the PolyForm license file from the repo and the release assets (#113)

### LINKED ISSUE #113: Remove LICENSE-PolyForm-Noncommercial-1.0.0.md from the repo and the release assets
Chris's direct order, 2026-09-08. The repo and every release asset carry `LICENSE-PolyForm-Noncommercial-1.0.0.md`. It was retained in 0.14.2 so a download could be compared against the terms it replaced, but it is not the license base ships under (FSL-1.1-ALv2 since 0.14.2) and a reader who opens it first concludes base is noncommercial-only and leaves.

## What to do

1. Delete `LICENSE-PolyForm-Noncommercial-1.0.0.md` from the tree.
2. Remove it from the release packaging: `.github/workflows/release.yml` copies it into all four assets, and `scripts/test-release-packaging.sh` plus the 0.14.2 assets check count four license files. Every reference must go in the same PR or CI reds on the asset count.
3. `LICENSING.md` keeps ONE sentence of history: versions up to and including 0.14.1 shipped under PolyForm Noncommercial 1.0.0, linking to the license text upstream (polyformproject.org) rather than hosting it. Those versions keep their terms; nothing is withdrawn.
4. Census every mention first (`git grep -n PolyForm`), including README, LICENSE.md, Cargo.toml metadata and the CHANGELOG. CHANGELOG entries are history and stay as written.

## Acceptance

- `git grep -n PolyForm` on the PR head returns only CHANGELOG lines and the one LICENSING.md history sentence.
- The packaging test asserts THREE license files and reds on the current main (law 25) before the change lands.
- Rides in the rank:03 CI/packaging lane.

<sub>Filed by the maintainer's automation (Claude Code, session auk) on his behalf.</sub>

ChristopherKahler:
Closes #113.

base has shipped under FSL-1.1-ALv2 since 0.14.2. `LICENSE-PolyForm-Noncommercial-1.0.0.md` was
kept in the tree and packed into all four release assets so a downloader could compare a version
against the terms it replaced. That has outlived its purpose: a reader who opens it first
concludes base is noncommercial-only and leaves.

## What moved, and why all of it had to move together

| file | change |
|---|---|
| `LICENSE-PolyForm-Noncommercial-1.0.0.md` | deleted (133 lines) |
| `.github/workflows/release.yml` | dropped from the Unix `cp` list, the Unix `tar` **member list**, and the Windows `Copy-Item` list; the two comments that explained why it was packed are rewritten |
| `scripts/test-release-packaging.sh` | `LICENSES` asserts three files, not four |
| `LICENSING.md` | one history sentence, now linking upstream |

The `tar` member list is a separate edit from the `cp`: `tar` packs the members it is named, so
a file removed from one and not the other leaves the two archives disagreeing. That asymmetry is
the bug `test-release-packaging.sh` was written for, and it is why the packaging gate reds if any
one of these three is missed.

## Kept, because history is not drift

Per the round's ruling on this issue, a PolyForm reference is drift only where it describes
**current** licensing or **current** packaging. These describe neither:

- **`CHANGELOG.md`** — what 0.14.2 said when it shipped.
- **`scripts/changelog-notes/0.14.2.md`** — the source `changelog.py` extracts for that version.
- **`verification/base-0.14.2/release_license_assets_0142.sh`** — the frozen acceptance harness
  for the change that *added* the fourth file. It is pinned to `BASE_SHA=f7c4f43`, it verifies
  the sha256 of the published v0.14.1 assets against the GitHub API, and it is not wired into CI.
  Rewriting a frozen instrument to describe a later tree makes it a false record of what that
  round actually verified, which is worse than a stale grep result.
- **`LICENSING.md`** — one sentence.

That last one is a **rewrite, not a keep**: the sentence linked the *local* file, which no longer
exists, and #113 asks for the upstream text. It now reads:

> Every release up to and including **v0.14.1** was published under the
> [PolyForm Noncommercial License 1.0.0](https://polyformproject.org/licenses/noncommercial/1.0.0).

The two sentences after it are untouched, and they are the part that matters: *that grant is not
withdrawn and cannot be*, and *if you received one of those versions, you keep the rights you
were given for that version, permanently.*

## A correction to #113's census

The issue says to census *"including README, LICENSE.md, Cargo.toml metadata and the CHANGELOG"*.
**None of those first three contains a PolyForm mention at all** — `Cargo.toml:5` already reads
`license = "FSL-1.1-ALv2"`. The real census is 15 lines across 7 files, all of them in the tables
above. Saying so is the point: "check README" otherwise reads as an open item forever.

The acceptance grep here is `git grep -ni polyform`, case-**in**sensitive, which is stricter than
the one #113 specifies. It catches `polyformproject.org`, which the case-sensitive grep misses.

## Acceptance

Run on Windows Git Bash with `pwsh 7.6.5` present, so the packaging test's Windows leg **genuinely
runs**. That matters: without `pwsh` that leg takes its `skip` branch, and a skip that is read as
a pass would mean half this change was never checked.

**RED, on unmodified `main` @ `610636e8`** — pointing `test-release-packaging.sh` at three
licenses while the workflow still names four:

```
PolyForm mentions in the test: 2 -> 1
exit 1, 18 ok / 1 FAIL
  FAIL  the step exited 2
  FAIL -- 1 check(s) above. A release cut from this commit would ship assets without them.
```

It fails **inside the packing step**, which is the right place: `cp` cannot find the fourth file,
so no archive is produced at all. A red in the root-file loop instead would have meant something
else was wrong.

**GREEN, on this branch @ `d159bef`** — 10 of 10 rows:

```
P2  packaging test exit 0, 19 ok / 0 FAIL / 0 skip
    the Windows leg RAN (the step ran under 7.6.5)
    LICENSE.md          asserted in 2 of 2 archives
    LICENSING.md        asserted in 2 of 2 archives
    LICENSE-APACHE-2.0  asserted in 2 of 2 archives   -> 6 assertions actually ran
P3  MASK: licenses kept in the cp, stripped from the tar member list -> exit 1, 3 FAIL
P4  census: 4 files, all in the history set; LICENSING.md has exactly 1 line;
    it links upstream; it no longer links the deleted local file
P5  not on disk, and not in `git ls-files` -- deleted, not emptied
```

**P3 is the leg that matters most**, and it is the same mutation
`verification/base-0.14.2/release_license_assets_0142.sh:245` used on this seam: deleting the
`LICENSES` loop outright would pass P2 identically while leaving the archives unguarded. P3 keeps
the `cp` and strips the `tar` member list, and the gate must still red. It does.

P4 prints the files it **visited**, and zero visited is a failure rather than a pass — a census
that matched nothing would otherwise look like a clean sweep.


ChristopherKahler:
Merged by auk on raven's G2 verdict (CONFIRMED at head `d159bef`, sha re-verified unchanged immediately before the merge) plus CI green read at STEP level, run 34222761096: release-invocations steps 3/4/5 success (`test-release-packaging.sh` is the very gate this PR changes), clippy step[5] `cargo clippy --all-targets -- -D warnings` success, test step[5] `cargo test` + step[6] release rehearsal success.

**Acceptance is the NARROWED one (round amendment A3), not #113's original line.** #113 as written asked that `git grep -n PolyForm` return only CHANGELOG lines and one LICENSING.md sentence. That was unsatisfiable and was confirmed so by three independent sweeps. The acceptance now reads: no PolyForm reference survives in any file describing CURRENT licensing or CURRENT packaging. Kept as history: `CHANGELOG.md`, `scripts/changelog-notes/0.14.2.md`, everything under `verification/base-0.14.2/`, and one `LICENSING.md` sentence. Editing a frozen instrument pinned to `BASE_SHA=f7c4f43` would rewrite the record of what 0.14.2 actually verified, which is worse than a stale grep result. History is not drift.

**raven's independent verification, beyond rerunning the builder's legs:**
- Red-first re-run on TODAY's main (745b1d04) rather than citing the handoff: `raven_g2_polyform.sh` (md5 55007d69) -> rc 5, all five checks red with the 8 offending lines named. PR head -> rc 0, twice.
- `scripts/test-release-packaging.sh` at the head: exit 0, 19 ok / 0 FAIL / 0 SKIP. The Windows leg GENUINELY RAN under pwsh 7.6.5 -- the row reads that it ran, not the skip branch, because a skip is not a pass. Both archives built and read: 14 tar members, 12 zip members, all three licenses asserted in BOTH.
- **Three mask legs raven added that the builder had not tested.** M1 is the one that mattered: dropping `LICENSING.md` from the Windows `Copy-Item` while leaving Unix untouched -> rc 1, and the zip does not carry it. **The Windows half of this guard had never been mutated by anyone.** M2 (Unix tar list drops `LICENSE-APACHE-2.0`) -> rc 1. M3 (license deleted from the tree) -> rc 1, refuses to rehearse. Each printed its diffstat first, so a mutation that edited nothing would report VOID rather than pass. Tree restored clean after each.
- Census: 7 surviving PolyForm lines, every one inside A3's kept set, ZERO offending. `verification/` untouched by the PR, as A3 requires. Deletion is real in git (`ls-files` 0, `diff-filter=D` names it). `LICENSING.md` links upstream, and BOTH comments were rewritten with the code so no stale "four files" prose survives -- that last one is the law-19 shape, a doc sentence outliving the code it described.

PR #124, opened 2026-09-08
Test and lint on Windows (#108)

### LINKED ISSUE #108: ci: there is no Windows job - base ships a Windows binary that is never tested and never linted

ChristopherKahler:
Closes #108.

base publishes `base-windows-x86_64.zip` on every release. **No test and no lint has ever run on
Windows.** `release.yml`'s Windows matrix leg is a plain `cargo build --release` (`:60`) — compiled
once, zipped, shipped. The only `cargo test` in that workflow is the ubuntu docs gate at `:22`, and
`ci.yml` had no `windows-latest` anywhere.

## Two jobs, separate on purpose

**`test-windows`** runs `cargo test` on `windows-latest`. Not a matrix leg on `test`: that job's
second step is `./scripts/test-release-rehearsal.sh`, a bash release rehearsal that is Linux-shaped,
so a Windows leg would need an `if:` skipping it. A job named `test` that quietly runs half of what
its name says is worse than no job, because it reports a name it did not earn.

It pins Python. `ast_extension_registry.rs`, `ast_automap_test.rs` and
`ast_extractor_notices_test.rs` shell `python` on Windows and **fail rather than skip** when it is
absent, by their own design — *"a skip here is indistinguishable from a pass"*. Trusting the image
to ship an interpreter would make a real failure and a missing interpreter look alike.

**`clippy` gains an os matrix**, `fail-fast: false`. A Linux-only lint is how `doorbell.rs`'s
`BUDGET` — dead on Windows and only on Windows — passed every check base had for months (#97, fixed
in #119). `fail-fast` off so a failure on one platform never hides the other's result; telling the
two apart is the entire point of adding the second platform.

## This PR opened RED deliberately, and the measurement corrected the prediction

#108 reports `tests/hook_log_tier_test.rs` failing **4 of 4** on Windows while passing on ubuntu,
and says plainly that its author did not re-run it and that *"the root cause is not established in
this issue"*. This PR landed the job first and measured it on GitHub's own runner rather than
designing a fix by reading `cfg` attributes.

**The measurement did not find what the issue predicted.**

### Run `34223265443`, at step level

| job | conclusion | what the **log** says |
|---|---|---|
| `clippy (ubuntu-latest)` | success | |
| `clippy (windows-latest)` | **success** | 248 `Checking`/`Compiling` crate lines; `Checking base v0.14.2` present as the **final** crate; `Finished dev profile in 8m 00s`; **0** `warning:` lines |
| `test` | success | |
| `release-invocations` | success | |
| `test-windows` | **failure** | `ast_automap_test` — 12 passed, **2 failed** |

### `hook_log_tier_test` never ran, so #108's own claim is still unmeasured

Occurrences of `hook_log_tier` in the entire 1468-line job log: **zero**.

`cargo test` aborts at the first failing target. It reached four binaries — `src/lib.rs`,
`src/main.rs`, `tests/analyze_catchall_test.rs`, `tests/ast_automap_test.rs` — and stopped there:

```
error: test failed, to rerun pass `--test ast_automap_test`
```

`ast_…` sorts before `hook_…`, so the file the issue was filed about never executed. Neither
confirmed nor refuted; **nobody has measured it yet**, including this run.

### The failure that is actually blocking, in `tests/ast_automap_test.rs`

```
---- the_temp_rule_stands_down_only_inside_a_redirected_home ----
tests\ast_automap_test.rs:281:9
assertion `left == right` failed: a fixture inside the sandbox home is mappable
  left: Some("a Windows AppData directory")
 right: None

---- the_sandbox_exemption_covers_the_segments_the_sandbox_itself_sits_under ----
tests\ast_automap_test.rs:318:9
assertion `left == right` failed
  left: Some("a Windows AppData directory")
 right: Some("a node_modules tree")
```

Both return the same reason, which means `sandbox_home()` did not recognise the sandbox and
`never_map_reason` read the whole path instead of the fixture-relative part.

Note what these two tests are: `ast_automap_test.rs:296-301` records that this exact shape was
**already measured once** — *"5 of 13 tests in this file failed on the Windows mirror at `b72703a`
for exactly this reason (2026-09-03); Linux never saw it"* — and `automap.rs:746-752` documents the
Windows behaviour as intended. Root cause is being established by direct instrumentation before any
fix is written or any scope call is made.

### `--no-fail-fast`, added in this PR

The first run is the argument for it: it stopped at the fourth binary and never executed the file
#108 was filed about, so **the job built to measure #108's claim did not measure it**. A first-ever
Windows job that reveals one failure per eight-minute round trip is worth little. The job now
reports every Windows failure in one run.

### `clippy (windows-latest)`: #119's undischarged MSVC leg is discharged

#119's merge comment recorded its MSVC corroboration as **undischarged** — it could not be run
locally at 2.0 GB free RAM. This leg is that instrument, and it was read at step level from the
**log**, not from the conclusion: a green exit alone cannot exclude a lint that ran on nothing.
It compiled 248 crates and linted `base` itself, last of the 248, with zero warnings.

https://claude.ai/code/session_011VSEdejdH235vTHgN3g6Rz

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,14 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.12'
-      - run: cargo test
+      # --no-fail-fast because cargo test ABORTS at the first failing target and
+      # runs the rest of them never. This job's first run proved the cost: it
+      # stopped at ast_automap_test (4th of the suite, alphabetically) and so
+      # tests/hook_log_tier_test.rs -- the failure #108 was FILED for -- did not
+      # execute at all, and #108's own claim went unmeasured by the job built to
+      # measure it. A first-ever Windows job that reveals one failure per 8-minute
+      # round trip is worth little; this one reports every Windows failure at once.
+      - run: cargo test --no-fail-fast
 
   # Required, not advisory. It was `continue-on-error: true` "until the three
   # pre-existing never_loop denies in src/crud are cleared" -- they are: a
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -299,7 +299,7 @@ jobs:
   # TEMPORARY. REMOVED BEFORE G2. Do not merge with it.
   #
   # R1 EXTENSION. The first experiment measured the release profile against ONE
-  # of the four targets #129 covers (hook_log_tier_test). R1's action gates only
+  # of the spawning targets #129 covers (hook_log_tier_test). R1's action gates only
   # the 2 ast_automap tests and moves the Windows job to --release, which leaves
   # 14 tests ungated on one target's evidence. If the other three still overflow
   # in release, that job's first full run turns main RED for every later PR.
@@ -309,7 +309,7 @@ jobs:
   # overflows" is established only for hook_log_tier_test, and assuming it for
   # the rest would be the same unmeasured leap in the opposite direction.
   #
-  # The job-level verdict is the STRICTEST of the four, never a majority.
+  # The job-level verdict is the STRICTEST of the set, never a majority.
   # ==========================================================================
   windows-release-all-targets-EXPERIMENT:
     runs-on: windows-latest
@@ -333,12 +333,12 @@ jobs:
           ls -l target/debug/base.exe   2>&1 || echo "NO DEBUG BINARY"
           ls -l target/release/base.exe 2>&1 || echo "NO RELEASE BINARY"
 
-      - name: Four targets x two profiles, rc captured to files before any reporting
+      - name: Five targets x two profiles, rc captured to files before any reporting
         if: always()
         shell: bash
         run: |
           set +e
-          for t in hook_log_tier_test hook_output_channel_test outside_workspace_test served_count_coach_test; do
+          for t in hook_log_tier_test hook_output_channel_test outside_workspace_test served_count_coach_test ast_relation_query_test; do
             cargo test --test "$t" -- --nocapture > "dbg_$t.log" 2>&1
             echo "$?" > "dbg_$t.rc"
             cargo test --release --test "$t" -- --nocapture > "rel_$t.log" 2>&1
@@ -347,7 +347,7 @@ jobs:
           done
           set -e
 
-      - name: PER-TARGET VERDICT (same pre-registered ladder, applied four times)
+      - name: PER-TARGET VERDICT (same pre-registered ladder, applied per target)
         if: always()
         shell: bash
         run: |
@@ -359,7 +359,7 @@ jobs:
           rank () { case "$1" in V1) echo 5;; V2) echo 4;; R2) echo 3;; R3) echo 2;; R1) echo 1;; esac; }
 
           echo "=================== PER-TARGET FACTS ==================="
-          for t in hook_log_tier_test hook_output_channel_test outside_workspace_test served_count_coach_test; do
+          for t in hook_log_tier_test hook_output_channel_test outside_workspace_test served_count_coach_test ast_relation_query_test; do
             tr -d '\r' < "dbg_$t.log" > "dbg_$t.clean" 2>/dev/null || : > "dbg_$t.clean"
             tr -d '\r' < "rel_$t.log" > "rel_$t.clean" 2>/dev/null || : > "rel_$t.clean"
             d_rc=$(cat "dbg_$t.rc" 2>/dev/null || echo NO_RC)
@@ -382,9 +382,9 @@ jobs:
           done
 
           echo "=================== JOB VERDICT ==================="
-          echo "STRICTEST_ACROSS_FOUR_TARGETS=$worst"
+          echo "STRICTEST_ACROSS_TARGETS=$worst"
           if [ "$worst" = "R1" ]; then
-            echo "All four targets read R1. The release profile is clean on every target"
+            echo "All targets read R1. The release profile is clean on every target"
             echo "the R1 action decides the fate of. Merging on it is now a measurement."
           else
             echo "NOT R1 across the set. One or more targets denies it, so the R1 action"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,14 +99,48 @@ jobs:
       # this runner is ubuntu — so the check reads `git check-attr`, not the checkout.
       - run: ./scripts/test-crlf-attributes.sh
 
+  # base publishes base-windows-x86_64.zip on every release and, until 0.14.3, no
+  # test and no lint had ever run on Windows (#108). release.yml's Windows leg is a
+  # plain `cargo build --release`: compiled once, zipped, shipped, and nothing else
+  # on that platform ever checked.
+  #
+  # Its own job rather than a matrix leg on `test`. That job's second step is
+  # ./scripts/test-release-rehearsal.sh, a bash release rehearsal that is
+  # Linux-shaped, so a Windows leg would need an `if:` skipping it -- and a leg
+  # called `test` that quietly runs half of what the name says is worse than no
+  # leg at all, because it reports a name it did not earn.
+  test-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      # ast_extension_registry.rs, ast_automap_test.rs and
+      # ast_extractor_notices_test.rs shell `python` on Windows and FAIL rather
+      # than skip when it is absent (ast_extension_registry.rs:14-15, :33 -- "a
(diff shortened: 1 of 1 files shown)


ChristopherKahler:
**Closing unmerged. The lane is killed, and the experiment this PR carried has already paid out.**

### What the experiment measured

Deciding run **34249727125**, job **102140770201** — one job, one runner, two arms differing only in the cargo profile. `BUILD_DEBUG_RC=0 BUILD_RELEASE_RC=0`, every arm ran 3–4 tests.

| target | debug rc / ovf | release rc / ovf |
|---|---|---|
| `hook_log_tier_test` | 101 / 4 | **0 / 0** |
| `hook_output_channel_test` | 101 / 5 | 101 / **0** |
| `outside_workspace_test` | 101 / 3 | **0 / 0** |
| `served_count_coach_test` | 101 / 4 | **0 / 0** |
| `ast_relation_query_test` | 101 / 6 | **0 / 0** |

Every debug arm reproduced; **`ovf=0` in every release arm**. The single target that denies a clean result is `hook_output_channel_test`, and it is not a #129 casualty at all: it reads `src/hook/mod.rs` off disk as a string and asserts on its text, with no `CARGO_BIN_EXE` and no child process, so it cannot change behaviour with the profile. Its failure is a CRLF line-ending artifact of the runner — which is #125, closed `not_planned`.

**Failure attribution: #129 = 13, CRLF = 1, #128 = 2, total 16.** `ast_relation_query_test` is a newly identified #129 casualty target, taking the measured Windows set from 16 to about 20.

### Why this PR does not merge

`--no-fail-fast` was the one genuinely load-bearing thing in this branch, and **this PR never delivered it to the trunk.** Diffing `ci.yml` on `main` against this head: #124 adds a new `test-windows` job, turns `clippy` into a two-OS matrix, and adds two temporary experiment jobs — **it never touches the ubuntu `test` job**, which is where the masking defect lived. That fix landed separately in #135 and is on `main` now.

What remains here is a Windows CI job whose value is gated on #129, which is measured non-user-facing and deferred. Under the standing triage bar — does this change whether base works for someone using it — it does not.

`tern/108-windows-ci-job` is **kept**, not deleted. #108 stays open and deferred; reopen this work when #129 is worth paying for.

Issue #121, opened 2026-09-08
Lock tripwire's session_start.rs "sentinel" is an exemption: it guarantees the tripwire will NOT fire there


ChristopherKahler:
### base version

base 0.14.2 (Cargo.toml `version = "0.14.2"`, measured on `610636e`)

### Host and OS

Claude Code on WSL or Linux

### What happens

`tests/lock_tripwire_test.rs` lists `src/hook/session_start.rs` in its allow-list, and both the
test's own reason string and the fork doc describe that entry as a **sentinel**: the fork doc's
Definition of Done says it is *"listed so the tripwire fires if one ever appears, in the file where
an unlocked write would be worst because session start runs on every session of every workspace"*,
and adds *"do not delete it as dead weight."*

**The entry does the exact opposite.** `ALLOW` / `ALLOW_FILES` is an EXEMPTION list, checked with
`if allowed.contains(rel) { continue }`. Being on it guarantees the tripwire **never fires** for
that file. A new unlocked graph writer in `session_start.rs` is silently permitted, and the entry
that is supposed to catch it is what permits it.

Measured as a pair, so the cause is not inferred:

| arm | injected unlocked `write_back` in `session_start.rs` | `ALLOW_FILES` entry | cargo rc | site named |
|---|---|---|---|---|
| M2 | yes | **present** | **0 (GREEN)** | 0 times |
| control | yes | removed | 101 (RED) | 1: `src/hook/session_start.rs::mutation_probe_unlocked_graph_write  [rule 1, write call: write_back(]` |

The control is what makes this evidence rather than a guess (law 24): with the entry removed the
rule *does* catch the very same injection, so the green in M2 is caused by the exemption and not by
the rule failing to see the code. Both arms compiled and both ran the test binary — checked, so
neither is a void result.

### How to reproduce

From a clone at `610636e` (or this branch):

```
# 1. add an unlocked graph write to the sentinel file, ABOVE its #[cfg(test)] line
#    (insert immediately before the `#[cfg(test)]` at src/hook/session_start.rs:592)
fn mutation_probe_unlocked_graph_write(store: &oxigraph::store::Store, path: &std::path::Path) {
    let _ = crate::store::write_back(store, path, crate::changelog::Change::Op("probe"));
}

# 2. run the tripwire
cargo test --test lock_tripwire_test -- --nocapture
#    -> test result: ok. PASSES. The new unlocked writer is not mentioned.

# 3. now delete this one line from ALLOW_FILES and run again
[...]

PR #126, opened 2026-09-08
test(lock): widen the tripwire to filesystem primitives, red-first (#104)

### LINKED ISSUE #104: Lock tripwire matches write call names, so graph_move.rs and doctor.rs::restore_tier are invisible to it
## The lock tripwire matches write CALL NAMES, so two production graph-write routes are invisible to it

Found by **shrike** while censusing #87. Read off `origin/main` @ `4ee4a455`.

`tests/lock_tripwire_test.rs` recognises exactly three write tokens — `write_back`,
`update_and_write`, `mutate_and_write` — and exempts whole FILES via a 12-entry ALLOW array.
Anything that writes a graph by another route has never once been looked at.

### Two such routes, both production

**4A — `src/graph_move.rs`.** A complete, independent write path sharing no code with
`store::write_back`. Zero lock tokens in the entire file.

| fn | line | what it does |
|---|---|---|
| `write_validated` | 469-486 | `fs::write(tmp)` → graph_health check → `fs::rename(tmp, path)` |
| `commit` | 438-464 | writes **both** `spec.dest_path` and `spec.source_path` |
| `rollback` | 451, 454 | `fs::copy(backup, source_path)` and `copy(b, dest_path)` |

Not on the ALLOW list, and it never needed to be — it uses none of the three tokens. Reachable
from `cli.rs:3965` (`base graph move`) **and from `crud/project.rs:627`**, so it is on a
user-facing project CRUD path, not an admin-only command. Two graph files written per call; the
failure path copies snapshots over both.

**4B — `src/doctor.rs:909-911`, `restore_tier`.** `fs::copy(backup, tmp)` then
`fs::rename(tmp, path)`. Whole-file overwrite, no lock, none of the three tokens.
Called from `cli.rs:3762`.

### Why 4B has to be fixed BEFORE #87, not with it

`doctor.rs` is on the ALLOW list today, so it is exempt regardless. #87 removes that exemption and
locks `repair_tier` at `:793`. **The tripwire then reports doctor.rs GREEN** — while `restore_tier`
at `:909` is still overwriting the graph unlocked, because the tripwire cannot see `fs::rename`.

We would ship #87, watch the instrument go green on the very file we just changed, close the issue,
and leave an unlocked whole-file graph overwrite behind — in the file we just certified. That is
not #87 failing. That is #87 succeeding and the certificate being worthless.

It is law 24 in a new costume: **a gate whose blind spot and whose pass look identical from
outside.**

### Proposed detection rule (shrike's)

Stop matching write **call names**. Flag any function that names a graph path and calls a rename or
copy primitive without a lock token. Name-matching is precisely why `graph_move.rs` was never
examined.

### Acceptance leg — and it EXPIRES when #87 merges

The widened tripwire must **flag `graph_move.rs` and `doctor.rs::restore_tier` on the current
tree**. It has to go **RED today**. A version that is green today detects nothing and would swap one
blind instrument for another (law 25: the leg must include the failure the fix could have masked).

Once #87 lands and locks those sites, that leg is no longer available. This is why the order is
fixed: **read-back after write → widen the tripwire → #87 itself**, each verified by the one before
it. Shipping the instrument change inside #87 would make #87 grade its own homework.

### Method, stated honestly

shrike did not grep the three tokens. It swept all 87 `.rs` files at `4ee4a455` for every filesystem
write primitive — `fs::write`, `File::create`, `OpenOptions`, `fs::rename`, `fs::copy`,
`dump_to_writer`, `serialize_to_write`, `BufWriter::new` — 210 primitive lines, intersected with
files naming a graph path.

**The oxigraph route is closed, measured.** `dump_to_writer`, `dump_graph_to_writer`,
`dump_dataset`, `RdfSerializer`, `bulk_loader`, `serialize_` swept across all of `src` at
`4ee4a455` **excluding `src/store.rs`**: **zero hits.** The only occurrence in the tree is
`store.rs:730`. So oxigraph's serializer is not reachable as a write route outside the seam.

That closes the argument rather than narrowing it: bytes reach a graph file by exactly two means —
oxigraph's serializer, or a `std::fs` primitive. Serializer outside the seam is zero, measured;
every `std::fs` primitive is enumerated. **Therefore there are three write routes and all three are
named.**

**The gap that remains, stated by shrike rather than discovered later:** this is a source-level
argument over `src` at one commit. It does not exclude a write via a shell-out or via a dependency,
and it is a grep over primitives, not a type-level guarantee. Within those bounds, complete.
Outside them, not a claim.

ChristopherKahler:
## What

The lock tripwire stops matching write CALL NAMES. A function that puts bytes over a path it names
as a graph must hold the graph lock, whatever filesystem primitive it reaches for.

Closes #104. **Instrument only, no production code moves here.** Locking the writers it now sees is
#87, next in this lane.

## Why the instrument had to be fixed before #87

`tests/lock_tripwire_test.rs` recognised three tokens, `write_back(`, `update_and_write(` and
`mutate_and_write(`, and skipped any file containing none of them *before examining a single
function*. `src/graph_move.rs` has its own complete write path (`fs::write` then `graph_health` then
`fs::rename`, plus rollback `fs::copy` over **both** graph files) and uses none of the three, so
**the tripwire had never opened that file.** Not "opened it and passed it". `doctor::restore_tier`
(`fs::copy` plus `fs::rename` straight over the graph) is the same shape behind a file-scoped
exemption.

Measured on `610636e` with the instrument unchanged:

```
tripwire: scanned 99 files under src/
tripwire: 10 graph-writing function(s) required to hold the lock
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
```

Green, while **twelve** unlocked whole-file graph writers sat in that tree: the nine `write_back`
sites behind file-scoped exemptions, and three fs-primitive routes it never opened a file to find.

That green is why #87 cannot be verified through the old instrument. #87 removes `doctor.rs`'s
exemption and locks `repair_tier`; the tripwire would then go green **on `doctor.rs`** while
`restore_tier` still overwrote the graph unlocked, and #87 would ship with a worthless certificate.

## Red-first (law 25)

The three legs land in the first commit against the **unwidened** detector, so their RED run is real
and reproducible rather than asserted afterwards. The second commit turns on exactly the enforcement
and nothing else.

Both arms below were re-run on **this** branch, on today's base `de01d7b`, after it was rebased
forward. The shas in the header are the ones the runner checked out and recorded in each log, and it
derives them from `git log` rather than taking them as arguments. The script that
opens this PR refuses if a log's recorded sha is not the commit it is about to label. That matters
because the earlier measurement was taken on commits that no longer exist on this branch, and
renaming a sha in a table would print a number for a commit that was never run.

| leg | `a748afa` red arm | `bf77dc1` widened |
|---|---|---|
| `the_widened_rule_sees_the_unlocked_fs_writers` (#104's acceptance) | **FAILED** | ok |
| `every_fs_primitive_shape_reaches_the_rule` (law 31 reach) | **FAILED** | ok |
| `every_graph_writer_outside_the_allow_list_takes_the_lock` | ok | ok |
| cargo | `FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s` | `ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.18s` |

The red arm is also the answer to "is the old detector still blind on today's main?". It runs the
new acceptance legs against the **unwidened** rule on base `de01d7b`, not on the older base the
first measurement used, and they still fail. The blindness is present tense.

The final head reproduces the green:

```
tripwire: 99 files scanned under src/
tripwire: 1277 non-test functions visited
tripwire: 626 test functions skipped
tripwire: 10 write-call function(s) required to hold the lock
tripwire: 157 fs write-primitive site(s) seen in non-test functions
tripwire: 9 unlocked function(s) name a graph path and call a primitive; 0 outside the rule 2 allow-list
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.17s
```

**The control hiding in those numbers:** write-call functions checked is **10** on the
final head, in the red arm and in the green arm alike, which is what the script that opens this PR
checks across all three logs before it will publish the claim. Adding the test-module skip therefore
introduced no rule-1 regression. Otherwise that would be prose.

## The three legs

* **`the_widened_rule_sees_the_unlocked_fs_writers`** is #104's acceptance leg. It runs the scan
  with **both allow-lists empty** and asserts the three known unlocked writers are reported **by
  name**, not by count. A rule too loose (exempts everything) and one too narrow (never fires) are
  both green in the main test; this is what separates them. It expires when #87 lands.
* **`every_fs_primitive_shape_reaches_the_rule`** is law 31. It proves the rule reaches every
  primitive shape **enumerated from the codebase**, and asserts the shape table equals
  `FS_PRIMITIVES` so the two cannot drift. Three negative controls prove it is not simply flagging
  everything: no graph path named, lock held, and a `#[cfg(test)]` fixture.
* **control counts.** A scan that visited zero files, zero functions or zero primitive sites proved
  NOTHING, and fails saying so rather than passing quietly.

`scan_file` takes both allow-lists as **parameters** rather than reading the constants. A detector
that can only be run with its own exemptions applied cannot be asked what it can *see*, which is the
question #104 is about, and it lets a reviewer re-run the scan with any lists and get finding
*sites* as strings.

## Two decisions worth reviewing

**Rule 2's exemptions are `(FILE, FUNCTION)` pairs, and that is load-bearing.** File scope is
exactly what would let `restore_tier` ship unlocked inside a `doctor.rs` the tripwire had just
certified green; repeating it in the wider rule rebuilds that hole one level up. Rule 1 keeps file
scope where it is right: `src/dashboard/api.rs` is one deliberate entry covering 17 sites.

**The test-module skip is working code now, not a comment.** The previous version carried *"Skip the
test modules: fixtures legitimately write graphs directly"* above code that skipped comment LINES
and no test module at all. Harmless while three named calls were matched; fatal with primitives in
scope, because **626 of the 1903 functions under `src/` are test code** (both counts
printed by the run above), including
`src/hook/session_start.rs`'s own `graph.nq` fixture, which is that file's only filesystem
primitive. Greening that by re-adding files to an allow-list would re-blind the exact files under
repair.

## What rule 2 flags today, and each disposition

9 functions, which is exactly the number of rows `ALLOW_FNS` declares, cross-checked
against the constant before this PR was opened. Every `ALLOW_FNS` reason names the non-graph file
that function actually writes: an exemption that cannot say what it writes is a hole.

| file::fn | what it writes | disposition |
|---|---|---|
| `graph_move.rs::write_validated` | the graph (temp then rename) | `(#87)`, row removed when it takes the lock |
| `graph_move.rs::commit` | rollback copies over **both** graphs | `(#87)`, same |
| `doctor.rs::restore_tier` | the graph (copy then rename) | `(#87)`, same |
| `store.rs::write_back_inner` | the graph, under the seam | permanent, this **is** the atomic write |
| `doctor.rs::repair_tier` | the `.quarantine-<stamp>` sidecar | permanent, its graph write is rule 1's |
| `graph.rs::auto_compact_tiers` | the `.last-auto-compact` marker | permanent, its graph write is rule 1's |
| `install.rs::create_global_tier` | `base.toml`, `standards.toml`, seed docs | permanent, writes no graph |
| `relay/mod.rs::export_nq` | `inbox.nq`, an ephemeral export | permanent, never a tier graph |
| `scaffold.rs::run` | `domains.toml`, `base.toml` | permanent, writes no graph |

9 entries now, 6 once #87 lands: the three `(#87)` rows come off together.

**Signal notes.** `"nq.` is not decoration: `doctor::restore_tier` is reached by **no other signal**,
its only tell being `path.with_extension("nq.tmp")`. And `lock_path(` is deliberately *not* a graph
signal: the relay has a `lock_path` of its own over its spool, and including it flagged
`relay::with_lock`, which touches no graph.

## Mutation check

Pulling the `graph_move.rs::write_validated` row out of `ALLOW_FNS` reds the suite and names it:

```
src/graph_move.rs::write_validated  [rule 2, fs primitive: "nq.]
```

## clippy, the gate and its control (law 24)

`rustc 1.95.0 (59807616e 2026-04-14)`

| run | rc |
|---|---|
| control, permissive `cargo clippy --all-targets` | **0** |
| gate, `cargo clippy --all-targets -- -D warnings` | **0** |

The permissive run is not decoration, and on this branch it did not stay theoretical. It caught two
things the gate alone could not have told apart.

**It separated a missing tool from a failing crate.** The first attempt at this leg returned rc 127
from a harness shell that had no `~/.cargo/bin` on PATH. Read through the gate alone that is just a
non-zero rc, indistinguishable from real warnings; the control named it as a missing tool, and no
verdict reached the record.

**It found a real red on this branch before the gate did.** The control returned rc 0 while printing
two `clippy::collapsible_if` warnings, both in this PR's own file:

```
tests/lock_tripwire_test.rs:367:9  this `if` statement can be collapsed
tests/lock_tripwire_test.rs:387:9  this `if` statement can be collapsed
warning: `base` (test "lock_tripwire_test") generated 2 warnings
```

Under `-D warnings` those became errors and the gate returned **101**, `could not compile base (test
"lock_tripwire_test") due to 2 previous errors`. `8272b67` collapses both with let-chains,
which is what clippy suggested and what `src/` already does in 80 places on edition 2024. Both sites
are pure conjunctions, no `else` on either arm and nothing between the outer `if` and the inner one,
so the collapse cannot move a printed count; the counts above are the re-run saying so rather than
this paragraph asserting it.

Control rc 0 proves clippy actually ran, which is what makes gate rc 0 mean something. Run against
`de01d7b`, after #119 made this a blocking check.

**The green log does not name what it covered, so the green was checked against the artifacts.** The
failing run named its target in the warning line; the passing run prints only `Checking base` and
`Finished`, which is a green that says nothing about whether the test target was linted at all. Two
fingerprints exist for `test-integration-test-lock_tripwire_test`, and they agree on `rustc`,
`target` and `path` and differ **only** in `profile`, which is the component that must differ
between a lint pass and a test build. The one written at 12:30:03Z falls inside the gate leg's
window of 12:29:50Z to 12:30:06Z; the other, at 12:30:09Z, is the tripwire building its binary
afterwards. That establishes the gate processed this file. It does not establish what a clippy
profile hash looks like in general, and is not offered as that.

## Found while testing this

**#121**, filed to milestone 0.14.4, unranked: the `session_start.rs` "sentinel" `ALLOW_FILES` entry
is an exemption, so it guarantees the tripwire will **not** fire there, the opposite of what it was
placed to do. Measured as a pair with a control, the same injected unlocked write: entry present
gives green and unnamed, entry removed gives red and named. Nothing is mis-permitted today; the
exposure is the future write the entry was placed to catch. Deliberately not fixed here.

`2bb3ac4` removes the words "the sentinel" from a doc comment in this PR's own file: the same
false capability claim, which I had written, about the instrument the claim is about.

---

Refs #104, #87, #121.

https://claude.ai/code/session_014PnkVcTBJ6PRiWAXFAUbhu

### DIFF
--- a/tests/lock_tripwire_test.rs
+++ b/tests/lock_tripwire_test.rs
@@ -274,9 +274,10 @@ fn rust_files(dir: &Path, out: &mut Vec<PathBuf>) {
 /// filesystem primitives in scope, because **626 of the 1,903 functions in
 /// `src/` are test code** and fixtures legitimately write graph files directly.
 /// `src/graph_move.rs` alone carries 8 `fs::write` calls past its
-/// `#[cfg(test)]`, and `src/hook/session_start.rs` — the sentinel — writes a
-/// `graph.nq` fixture. The cheap way to green that would be to re-add files to
-/// an allow-list, which re-blinds the exact files under repair.
+/// `#[cfg(test)]`, and `src/hook/session_start.rs` writes a `graph.nq` fixture
+/// which is that file's only filesystem primitive. The cheap way to green that
+/// would be to re-add files to an allow-list, which re-blinds the exact files
+/// under repair.
 fn functions(src: &str) -> Vec<Func> {
     let lines: Vec<&str> = src.lines().collect();
     let cfg_test_at = lines
--- a/tests/lock_tripwire_test.rs
+++ b/tests/lock_tripwire_test.rs
@@ -35,7 +35,8 @@
 //!   here: `src/dashboard/api.rs` is one deliberate entry covering 17 sites.
 //! * **Rule 2 — filesystem primitives.** A function that puts bytes over a path
 //!   it names as a graph must hold the lock, unless the (FILE, FUNCTION) pair is
-//!   on [`ALLOW_FNS`].
+//!   on [`ALLOW_FNS`]. Measured on `610636e`: 157 primitive sites in non-test
+//!   functions, 9 of them in a function that names a graph and takes no lock.
 //!
 //! **Rule 2's exemptions are function-scoped and that is load-bearing.** A
 //! file-scoped exemption is precisely what would have let `restore_tier` ship
@@ -377,10 +378,6 @@ fn scan_file(
 
         // Rule 2 — filesystem primitives over a path the function names as a
         // graph, exempted by (FILE, FUNCTION).
-        //
-        // NOT YET ENFORCED in this commit. The candidates are measured and
-        // PRINTED so the red-first run shows the detector naming the very sites
-        // it then fails to report as findings — the widening lands next.
         let sites = code
             .lines()
             .filter(|l| first_match(l, &FS_PRIMITIVES).is_some())
@@ -391,7 +388,11 @@ fn scan_file(
                 counts.fs_primitive_graph_functions += 1;
                 if !allow_fns.contains(&(rel, f.name.as_str())) {
                     counts.fs_primitive_functions_flagged += 1;
-                    println!("  rule 2 candidate (not yet enforced): {site}  [{sig}]");
+                    found.push(Finding {
+                        site,
+                        rule: Rule::FsPrimitive,
+                        token: sig.to_string(),
+                    });
                 }
             }
         }
--- a/tests/lock_tripwire_test.rs
+++ b/tests/lock_tripwire_test.rs
@@ -13,15 +13,55 @@
 //! writer with no test would sail through. Reading the source catches it whether
 //! or not anything calls it.
 //!
-//! Adding a site to the allow-list is a deliberate act that shows up in review.
-//! The list is PRINTED on every run, so an exemption is visible rather than
+//! Adding a site to an allow-list is a deliberate act that shows up in review.
+//! Both lists are PRINTED on every run, so an exemption is visible rather than
 //! assumed.
+//!
+//! # Two rules, because matching call NAMES was itself the blind spot (#104)
+//!
+//! The first version of this test recognised three write calls and nothing else.
+//! `src/graph_move.rs` uses none of them — it has its own `fs::write` →
+//! `fs::rename` path — so the tripwire **never opened that file**. Not "opened it
+//! and passed it": the file-level filter skipped it before any function was
+//! examined. `doctor::restore_tier` is the same shape behind a file-scoped
+//! exemption. Measured on `610636e`, the unwidened test scanned 99 files, checked
+//! 10 functions and reported all-clear while TWELVE unlocked whole-file graph
+//! writers sat in the tree.
+//!
+//! So:
+//!
+//! * **Rule 1 — write calls.** A function reaching the store seam's write must
+//!   hold the lock, unless its FILE is on [`ALLOW_FILES`]. File scope is right
+//!   here: `src/dashboard/api.rs` is one deliberate entry covering 17 sites.
+//! * **Rule 2 — filesystem primitives.** A function that puts bytes over a path
+//!   it names as a graph must hold the lock, unless the (FILE, FUNCTION) pair is
+//!   on [`ALLOW_FNS`].
+//!
+//! **Rule 2's exemptions are function-scoped and that is load-bearing.** A
+//! file-scoped exemption is precisely what would have let `restore_tier` ship
+//! unlocked inside `doctor.rs` while the tripwire certified that same file green.
+//! Repeating file scope in the wider rule would rebuild the hole one level up.
 
 use std::collections::BTreeSet;
 use std::path::{Path, PathBuf};
 
-/// Calls that write a graph file back to disk.
+/// Calls that write a graph file back to disk through the store seam.
 const WRITE_CALLS: [&str; 3] = ["write_back(", "update_and_write(", "mutate_and_write("];
+
+/// Filesystem primitives that can put bytes over a graph file WITHOUT touching
+/// the seam. Enumerated from the codebase, not from imagination: these are the
+/// primitives the #104 route sweep counted across `src/` (211 call-site lines at
+/// `610636e`), plus `fs::remove_file`, because deleting a graph is a whole-file
+/// write of zero bytes.
+const FS_PRIMITIVES: [&str; 7] = [
+    "fs::write(",
+    "fs::rename(",
+    "fs::copy(",
+    "fs::remove_file(",
+    "File::create(",
+    "OpenOptions",
+    "BufWriter::new(",
+];
 
 /// Anything that puts the graph lock in scope for the enclosing function.
 const LOCK_TOKENS: [&str; 5] = [
@@ -32,9 +72,34 @@ const LOCK_TOKENS: [&str; 5] = [
     "mutate_file_if_holds(",
 ];
 
-/// Files allowed to write a graph without holding the lock, each with the reason.
+/// Spellings by which a function names a graph FILE, for Rule 2.
+///
+/// `".nq"` covers every path literal (`graph.nq`, `dest.nq`). `"\"nq."` covers
+/// the temp-file spelling `with_extension("nq.tmp")` — and that one is not
+/// decoration: **`doctor::restore_tier`, 4B in #104, is reached by NO other
+/// signal.** Its only tell is `path.with_extension("nq.tmp")`; it names no
+/// graph-path variable and calls no graph function. Drop that entry and the
+/// widened rule goes blind to the exact site it was widened for.
+///
+/// The rest are the resolvers a graph path arrives through. `lock_path(`
+/// deliberately is NOT here: the relay has a `lock_path` of its own over its
+/// spool, and including it flagged `relay::with_lock`, which touches no graph.
+const GRAPH_SIGNALS: [&str; 10] = [
+    ".nq",
+    "\"nq.",
+    "graph_path",
+    "dest_path",
+    "source_path",
+    "all_tier_files",
+    "tier_paths(",
+    "graph_health(",
+    "load_graph(",
+    "load_or_empty(",
+];
+
+/// Rule 1: files allowed to reach the seam's write without holding the lock.
 /// Every entry is a deliberate exemption, not an oversight.
-const ALLOW: [(&str, &str); 12] = [
+const ALLOW_FILES: [(&str, &str); 12] = [
     ("src/store.rs", "the seam itself: it defines the lock and the write"),
     (
         "src/dashboard/api.rs",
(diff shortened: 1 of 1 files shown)


ChristopherKahler:
Merged at head `8272b6713a2dcdb275f6a339158960bf446faede` on raven's G2 plus CI green **read at step level**, not at job level.

## CI, step level, run `34226986066`

| job | substantive step | result |
|---|---|---|
| `clippy` | `cargo clippy --all-targets -- -D warnings` | success |
| `test` | `cargo test`, `release rehearsal` | success |
| `release-invocations` | invocations, packaging, crlf-attributes | success |
| `python-ast-tests` | `scripts/ast/test_*.py` | success |

No step in any job is anything but success or skipped.

## What settled the merge question

A tripwire that fails when an unlocked graph writer exists **shipped green**, and green had two readings
the run could not separate: (a) the widening reaches those writers and nothing is outside the allow-list,
or (b) the widening does not reach them and the green is a void. With 9 flagged and 9 allow-listed, (a)
and (b) produce identical output.

**(b) is excluded by measurement**, from raven's own red-arm log: the widened rule sees **0 of 3** known
unlocked fs-primitive graph writers before the widening and **3 of 3** at head. Same detector, same tree,
same target dir, minutes apart; the only variable is the widening. The discriminator is not circular — it
calls `scan_tree` with **both allow-lists empty** and asserts `functions_visited > 0` *before* the name
check, which is only possible because `allow_files`/`allow_fns` are **parameters, not constants**.

`every_graph_writer_outside_the_allow_list_takes_the_lock` is green in **both** arms, so the red arm fails
on the widening's absence and not on a broken harness.

## UNDISCHARGED LEGS — recorded here rather than dropped

1. **Rule 2 has one implementation, independently coded but not independently derived.** raven
   re-implemented rule 2 in Python using *deliberately the same tokens*. That can catch a wrong
   **implementation** — and did not; 626 test functions skipped in both, the same 9 candidate names — but
   it **cannot catch a wrong RULE**. Nothing here validates the token list itself.
2. **The clippy-coverage evidence carries a declared limit.** The fingerprint decomposition establishes
   that the gate **processed** `lock_tripwire_test` (three components identical, only `profile` differing,
   one write inside the gate window). It does **not** establish what a clippy profile hash looks like in
   general. condor narrowed this claim itself after it had already been accepted at full strength.
3. **9 flagged and 9 allow-listed means the tripwire cannot go red on real drift for those nine until
   #87 lands.** The instrument is correct and currently has nothing to catch. #87 is what removes the
   allow-list; until then this PR ships an instrument whose value is forward-looking.

## Instrument faults surfaced during the grade, both the reviewer's own

- An independent three-name classifier returned **perfectly inverted** results — all three targets marked
  clean, the control marked as an offender — because it tested "names a graph path" by looking for
  `graph.nq` in the function **body**, while `write_validated` and `commit` take the path as a
  **parameter** and never spell it. The control leg is the only reason it did not ship as a false defect
  against correct work. It is kept in the instrument table as a record rather than deleted.
- A brace splitter with no string awareness ran past a function end on a line containing the literal
  `"sx={"`, producing a 10th candidate. raven **read the function** before reporting it (a pure marker
  check, no primitive, no graph path) rather than filing it. condor's 9 stands.

`open_pr.py`'s no-renamed-measurement enforcement is **verified by the reviewer**, not asserted by the
builder: three arms with rc read with no pipe — unmutated control `rc 0`; the red log relabelled with the
**real green sha** `rc 1`; the red arm made to **report green** `rc 1`. The builder had tested only the
weaker of the two checks, using an obviously-fake `0000...0000` sha, and asked for that to be said plainly.

Filed separately during this review: **#127** (no retry on the atomic rename in `write_back_inner`;
reachable on NTFS), 0.14.4, unranked.

Issue #129, opened 2026-09-08
base.exe stack-overflows under cargo test on windows-latest: one 0xC00000FD behind 14 of the 16 Windows test failures


ChristopherKahler:
## Measured, not read from code

The first Windows CI run in the project's history (PR #124, `--no-fail-fast`) shows `base.exe` exiting
with **`STATUS_STACK_OVERFLOW`** when spawned as a child process by the test suite.

| evidence | value |
|---|---|
| exit code | `-1073741571` = `0xC00000FD` = **`STATUS_STACK_OVERFLOW`** |
| arithmetic, not a guess | `-1073741571 + 2^32 = 3221225725 = 0xc00000fd`; the log also prints `ExitStatus(3221225725)` literally |
| `thread 'main' has overflowed its stack` | printed **9** times, in plain words |
| the `-1073741571` exit | recorded **7** times |

## One cause, four targets, 14 tests

| target | failing | mechanism |
|---|---|---|
| `hook_log_tier_test` | 4 | spawns `CARGO_BIN_EXE_base`, gets `-1073741571` |
| `hook_output_channel_test` | 4 | same |
| `outside_workspace_test` | 2 | same |
| `served_count_coach_test` | 4 | same |

All four spawn the binary the same way — `Command::new(env!("CARGO_BIN_EXE_base"))` with `BASE_HOME`
pointed at a fresh temp home and `BASE_NO_AUTO_UPDATE=1`. The binary dies before it can answer.

The Windows run's other 2 failures (`ast_automap_test`) are **not** this. They are pure-Rust calls into
`never_map_reason` that never spawn the binary, and they have their own issue: #128.

**Three counts that did not trivially reconcile:** 14 failing tests, 7 recorded `-1073741571`
exits, 9 `overflowed its stack` messages. They are reconciled in the amendment at the bottom of this
issue, measured per failure block rather than by counting signature strings across the log. One of the
two explanations guessed at here — a test spawning more than once — is the correct one, and it is now
measured rather than supposed.

## #108's claim is confirmed; its diagnosis is wrong

`hook_log_tier_test` really is **0 passed / 4 failed** on Windows, exactly as #108 filed it.

But the failure is **not** in hook cwd-tier logic. All four asserts die on

```
hooks fail open: left Some(-1073741571) right Some(0)
```

— the assertion that the hook process **exited zero**. The tier logic is never reached. Anyone fixing
#108 by reading `src/hook/mod.rs` would be fixing a file that never ran.

[...]

Issue #128, opened 2026-09-08
automap: is_under compares an 8.3 short path against its long form, because canonicalize only expands a path that exists


ChristopherKahler:
## Measured, not read from code

`is_under`'s `norm()` closure (`src/hook/automap.rs:895`, at `bfcc2276`) calls

```rust
let real = std::fs::canonicalize(p).unwrap_or_else(|_| p.to_path_buf());
```

`canonicalize` expands a Windows 8.3 short name **only for a path that exists**. On failure the closure
silently falls back to the raw spelling. So when one side of the comparison exists and the other does
not, the two sides are normalised to *different spellings of the same directory* and the string compare
at the end of `is_under` returns `false`:

- a path that **exists** → canonicalize resolves it → `c:/users/runneradmin/...`
- a path that does **not** exist yet → stays as given → `c:/users/runner~1/...`

The GitHub Actions `windows-latest` runner's `std::env::temp_dir()` yields the **8.3 short form**, which
is what puts both spellings in play at once.

### The doc comment already claims this case is handled

Directly above the defect, `src/hook/automap.rs:888-892`:

> Both sides are canonicalised when that succeeds — which is what makes `%TEMP%`'s 8.3 short form
> `C:\Users\CHRIS~1\AppData\Local\Temp` match the long spelling — and compared raw when it does not,
> since a candidate root need not exist yet.

Both halves of that sentence are true and the conjunction is the bug: the 8.3 case is handled **only when
both sides exist**, and the "candidate root need not exist yet" allowance is exactly what reintroduces the
short form on one side.

## The two failing tests, and why they are one defect

`ast_automap_test` is 12 passed / **2 failed** on `windows-latest`. Both failures run through
`sandbox_home` (`src/hook/automap.rs:804`), whose entire body is one `&&`:

```rust
(under_any(&home, &temp_roots()).is_some() && is_under(root, &home)).then_some(home)
```

The two tests fail on **opposite operands of that `&&`**, and both operands are `is_under`:

| test | line | branch | evidence |
|---|---|---|---|
| `the_temp_rule_stands_down_only_inside_a_redirected_home` | :281 | **B** | `under_temp=None` while `is_under(root,home)=true`. The home **is** a temp path and `temp_roots()` did not recognise it. `home=C:\Users\RUNNER~1\AppData\Local\Temp\.tmpOu924F\home` |
[...]

Issue #125, opened 2026-09-08
`.gitattributes` does not pin `*.rs`, so a Windows-written Rust file commits with CRLF


ChristopherKahler:
`.gitattributes` pins `*.sh`, `*.py` and `*.ps1` with `text eol=lf`. It does **not** pin `*.rs` — `git check-attr` returns `text: unspecified` for every `.rs` file in the tree.

So a Rust file written by a Windows-side tool enters the blob byte-for-byte with CRLF intact, while an identically-written `.py` is normalised back to LF on the way in. Measured by osprey during the 0.14.3 round.

## The measurement

Census over every text blob at `745b1d0`: 160 `.rs`, 11 `.py`, 24 `.sh`, 32 `.md`, 5 `.yml` — **zero CRLF in any of them**. main has never carried this. On osprey's working branch, identical everywhere except `src/crud/ast_query.rs` at **897/897 CRLF**.

## Why it matters

The file read as **897 insertions / 721 deletions** under `git diff` and **176 insertions / 0 deletions** under `git diff -w`. The real change was 176 lines. A reviewer would have been handed a 1618-line diff for a 176-line change — and the natural response to a diff that size is to skim it.

## Note the doc predicted its own gap

The `.gitattributes` header's last paragraph reads: *"Leaving one script family unpinned in the commit that pins the rest is how the next gap gets created."* `*.rs` is that gap.

## Fix

Pin `*.rs text eol=lf`. Held out of osprey's 0.14.3 PR deliberately: it is a repo-wide bar change touching every lane's files, and round amendment A7 says bar changes go first and separately rather than riding a feature PR. The single existing casualty (`ast_query.rs`) is being normalised inside osprey's own rebase.

Filed to 0.14.4 unranked per the round's found-issue rule.

ChristopherKahler:
Closed as not planned.

Triage 2026-09-08 against one bar: does this change whether base works for someone using it. This is internal-instrument or developer-hygiene detail - the tool behaves identically for a user whether it is fixed or not. Reopen if a real user hits it.

ChristopherKahler:
## Reopened 2026-09-09 against a sharper bar

Chris's bar today: **anything that bites the building or the user needs resolving.** Edge cases and nice-to-haves stay closed. This one bites the building, with evidence gathered after the original triage.

| what | measurement |
|---|---|
[...]

PR #135, opened 2026-09-08
ci: --no-fail-fast on the trunk suite

ChristopherKahler:
One line. In `.github/workflows/ci.yml`, the ubuntu `test` job's `run: cargo test` becomes `run: cargo test --no-fail-fast`. Nothing else is touched.

## Why this one passes the bar

It is the instrument that decides whether real user-facing bugs are seen at all: while it is absent, one red target aborts the trunk suite and every target after it is never measured, so a genuine user-facing failure can sit on `main` reported as "one test failed". Verification discipline is explicitly exempt from the no-internal-work bar.

## The defect, measured

`cargo test` stops at the first test binary that fails. Run [34243724191](https://github.com/ChristopherKahler/base/actions/runs/34243724191) at `1199219a02b645bce6931e7e81e541fb1f46527c`, ubuntu `test` job `102120189361`, step `cargo test` = `failure`:

| | |
|---|---|
| integration targets in `tests/` at that head | **62** |
| targets the log shows `Running` | **4** |
| last target with a `FAILED` result | `ast_extension_registry` |
| targets that **never ran** | **58 of 62** |

cargo ran the two unit-test binaries and four integration targets in alphabetical order, hit a failure inside `ast_extension_registry`, and stopped. Among the 58 that never executed: `ast_relation_query_test`, `hook_log_tier_test`, `hook_output_channel_test`, `guard_test`, `lock_tripwire_test`, `fork_test`.

Those 58 are not green and not red on that commit. They are **unmeasured** — and the job reports one failure whether the true count is 1 or 40.

## What it does not change

- The job still fails when anything fails. `--no-fail-fast` changes what gets *reported*, never what gets *enforced*.
- No new jobs, no matrix, no runner changes, no test changes.
- Not a fix for #128 or #129. Those are separate and both measured non-user-facing this round.
- Deliberately does **not** touch `.gitattributes`. #125 was closed `not_planned`; nothing here reopens it.

## How the edit was made

Anchored on content, not on a line number: the exact byte string `        run: cargo test\r\n`, asserted to occur **exactly once** in the file, aborting without writing on any other count. This matters because an unanchored grep for `cargo test` in this file hits the step's *display name* first, and renaming a step passes review by eye while changing nothing.

The workflow was re-parsed after the edit and `cargo test --no-fail-fast` is present in the parsed `jobs.test.steps` structure, not merely in the text.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Cache cargo
         uses: Swatinem/rust-cache@v2
       - name: cargo test
-        run: cargo test
+        run: cargo test --no-fail-fast
       # Cut a whole release in a throwaway clone. It runs here rather than in a
       # job of its own because it reuses this one's target directory and cargo
       # registry; on its own it would pay for both again. 0.13.16's first


ChristopherKahler:
**G2 GRANTED** by `raven` at head `8d97714240b715ae2f93a5353b88ee73718f8302`, merged by `auk` with that sha pinned in the merge call.

### What justifies it — one-platform, same-run truncation, no confound

Run **34243724191**, ubuntu `test` job **102120189361**: there are **62 integration targets** in `tests/` at that head and **0 `[[test]]` stanzas** in `Cargo.toml`, so the filesystem is the denominator. **4 ran, 58 never ran.** The four that ran are the first four alphabetically and the fifth never started — that is fail-fast truncation at position 4, not selective running.

### Merging cannot redden `main`

`main`'s own `test` job **102137141797**: 64 `Running` lines (2 unittest binaries + 62 integration targets + 1 Doc-tests header) against 65 test-result lines, **0 FAILED, 881 tests executed**; headers and results reconcile exactly at 65. Under fail-fast a PASS means nothing was truncated, so all 62 targets already run green on `main` today. Adding the flag cannot reveal a hidden red here because there is no hidden red to reveal.

### Undischarged, stated rather than hidden

- **This PR run does not exercise the flag.** Both arms of run **34252526753** are green, so `--no-fail-fast` had no failure to continue past. That run proves the change does not BREAK the suite; it does not prove it WORKS. Not blocking — a `ci.yml`-only diff cannot make a green suite red.
- **No efficacy proof is claimed at all.** An earlier draft of this record cited a cross-platform pair as efficacy evidence; `raven` withdrew that sentence itself, for two faults — it was confounded (the arms differed in platform as well as in the flag, on a target that fails on Windows and passes on ubuntu) and it was misattributed. Whether a clean same-platform pair exists in #124's history is **explicitly unchecked**; #124 is being closed unmerged and that is archaeology under the round's standing bar.

PR #141, opened 2026-09-08
THROWAWAY red for the --skip-hooks regression in #136 — do not merge

ChristopherKahler:
**THROWAWAY. Do not review, do not merge.** Closed and its ref deleted as soon as the
result is read; it exists only to put a red on the record before the fix that clears it.

Head `e96401efe44fd184c6a7865cb71930215abb250c` is #136's head `806d2fa` plus
`tests/install_e2e_test.rs` and **nothing else** — no `src/` change of any kind.

The leg: `skip_hooks_holds_when_claude_code_is_already_installed`. A home where `~/.claude`
already exists (every machine that has Claude Code) and this version is unstamped, running
`base install --no-starter-commands --skip-hooks`, asserting `settings.json` **and** the
`.hooks-wired-<version>` stamp are both still absent.

Expected **RED on this head**: `cli::run` calls `ensure_hooks_wired()` before it dispatches and
excludes only `Commands::Hook`, so `install` reaches it too — the hooks are written and the
version stamped, and step 3's `⊘ Hook wiring skipped (--skip-hooks)` prints over work that has
already happened. Found by raven reading `src/cli.rs` and `src/install.rs`. The existing
`skip_hooks_still_means_skip_hooks_even_after_base_creates_the_directory` cannot reach it: its
home has no `~/.claude` at entry, so the seam returns empty and the leg passes for an unrelated
reason.

Everything else in the file and the suite is expected to stay green — a second failing leg here
would mean the new leg is wrong rather than the product.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -143,15 +143,18 @@
             Say '  then reopen your terminal.'
         }
 
-        # Hooks are only written into an existing ~\.claude. Installing base
-        # before Claude Code leaves it inert with no way to self-heal, because the
-        # repair runs from the session-start hook that was never wired. Test
-        # settings.json, not the directory: `base install` creates ~\.claude
-        # itself for the bundled skill, so the directory always exists afterwards.
+        # Since #93 `base install` wires the hooks even when Claude Code is not
+        # installed yet: it creates ~\.claude itself for the bundled skill, and
+        # the wiring is deferred to after that rather than dropped, so a later
+        # Claude Code finds them. This file therefore exists after every install
+        # that did not pass --skip-hooks; if it does not, the wiring FAILED
+        # rather than being deferred. Test settings.json, not the directory,
+        # which `base install` creates either way.
         if (-not (Test-Path (Join-Path $binHome '.claude\settings.json'))) {
             Say ''
-            Say 'base: no ~\.claude directory, so the hooks were not wired and base will'
-            Say '  not do anything yet. Install Claude Code, then run:'
+            Say 'base: the hooks were not wired, so base will not do anything yet.'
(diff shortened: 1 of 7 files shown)


ChristopherKahler:
**RED CONFIRMED for both legs. Closing — folded into #136.**

Run [`34258760997`](https://github.com/ChristopherKahler/base/actions/runs/34258760997), test job
`102171137400`, head `12c97d8` = #136's `806d2fa` plus `tests/install_e2e_test.rs` and nothing else.

Compiled clean (0 `error[E…]`, 0 `could not compile`), so this is a red and not a void. 66 targets
started (2 unittests + 64 integration), 67 `test result:` lines — targets plus the doc-test, which
reconciles exactly. Clippy green.

Attributed by test name through a map read from the tree, not by position in the log:

| target | visited | ok | FAILED |
|---|---|---|---|
| `install_e2e_test` | 8/8 | 6 | **2** |
| `install_scripts_source_test` | 7/7 | 7 | 0 |
| `hooks_self_heal_test` | 5/5 | 5 | 0 |

The whole-job census finds exactly those two names and no others:

- `skip_hooks_holds_when_claude_code_is_already_installed` — *"--skip-hooks wrote hooks anyway,
  into a home that already had ~/.claude."*
- `uninstall_does_not_create_a_settings_file_on_its_way_to_emptying_one` — *"uninstall created a
  settings.json that was never there."*

Green on the fix at `8504ddd`: [`34259230056`](https://github.com/ChristopherKahler/base/actions/runs/34259230056),
`install_e2e_test` 9/9 ok. Ref deleted.

PR #146, opened 2026-09-08
#87: the nine Tier C writers take the graph lock, and write_back goes private

### LINKED ISSUE #87: 10 graph writers still write unlocked (bulk/maintenance sites)

ChristopherKahler:
Closes #87 — **rank:01**, the highest-ranked open bug in the 0.14.3 round.

Built by `condor` on 2026-09-08 at `994ea29`, six commits, red-first. The branch has been
on the remote at this sha and has simply had no PR open, so the round's top-ranked item was
invisible in review all day. Opening it to correct that.

## What it does
The nine Tier C writers take the graph lock, `write_back` goes private, and
`migrate_trig_to_nq` takes the lock on a new bulk wait bound. A `LockedGraph` /
`lock_and_load_graph` pair plus a runtime identity backstop closes the seam.

## Commits
- `a46f33e4` red-first: narrow rule 1's `src/store.rs` file exemption to three named seam writers
- `7d3dca93` step 3 acceptance: five legs for the identity backstop, with a mutation arm
- `b9d44629` steps 2+3: `LockedGraph`, `lock_and_load_graph`, runtime identity backstop
- `6e3db59b` teach the tripwire the fix's lock spelling and the two graph-path names
- `c3510274` site ten: `migrate_trig_to_nq` takes the graph lock on a new bulk wait bound
- `994ea290` the nine Tier C writers take the lock, and `write_back` goes private

## State on opening — read this before grading
`git compare` against `main` at `0acd6e85` reports **diverged, 6 ahead / 6 behind**, 17 files.
The branch was rebased onto `3b736901`; main has moved six commits since. **It needs a rebase
onto current main before merge**, and CI on this head measures the older base.

Not yet graded. G2 has not been requested.

### DIFF
--- a/src/apply_ops.rs
+++ b/src/apply_ops.rs
@@ -445,27 +445,21 @@ pub fn run(graph_path: &Path, input: &str) -> (Value, i32) {
         return (Outcome::default().to_json(), 0);
     }
 
-    // A machine that has never written anything has no graph.nq yet, and a first
-    // pull is exactly that case — so start from an empty store rather than
-    // refusing, matching `crud::load_workspace_store`, the write path `base learn`
-    // already takes on a fresh home.
-    let store = if graph_path.exists() {
-        match crate::store::load_graph(graph_path) {
-            Ok(s) => s,
-            Err(e) => {
-                return (OpError::whole("graph_load_failed", e.to_string()).to_json(), 1);
-            }
-        }
-    } else {
-        match Store::new() {
-            Ok(s) => s,
-            Err(e) => {
-                return (OpError::whole("store_init_failed", e.to_string()).to_json(), 1);
-            }
+    // #87: lock, THEN load, and the conditional is gone rather than moved.
+    // `lock_and_load_graph` loads through `load_or_empty`, which already answers
+    // the case this branch existed for: a machine that has never written anything
+    // has no graph.nq yet, and a first pull is exactly that — an empty store, not
+    // a refusal. One route now covers both, so there is no second path to keep in
+    // step. The guard holds to the end of the function, covering the write below.
+    let locked = match crate::store::lock_and_load_graph(graph_path) {
+        Ok(g) => g,
+        Err(e) => {
+            return (OpError::whole("graph_load_failed", e.to_string()).to_json(), 1);
         }
     };
+    let store = locked.store();
 
-    let outcome = match apply(&store, &prepared) {
+    let outcome = match apply(store, &prepared) {
         Ok(o) => o,
         Err(e) => return (e.to_json(), 1),
     };
@@ -476,8 +470,7 @@ pub fn run(graph_path: &Path, input: &str) -> (Value, i32) {
         return (outcome.to_json(), 0);
     }
 
-    if let Err(e) = crate::store::write_back(&store, graph_path, Change::RemoteOps(&outcome.records))
-    {
+    if let Err(e) = locked.write(Change::RemoteOps(&outcome.records)) {
         return (OpError::whole("write_failed", e.to_string()).to_json(), 1);
     }
     (outcome.to_json(), 0)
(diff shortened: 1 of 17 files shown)


ChristopherKahler:
Closing: opened in error by `auk`.

This branch and sha (`994ea290`, `condor/87-lock-remaining-writers`) were **already merged as #139** at 2026-09-08T18:23:34Z. `main` and this branch are byte-identical on every marker checked — `LockedGraph` 15/15, `lock_and_load_graph` 2/2, `migrate_trig_to_nq` 4/4, and `tests/lock_identity_test.rs` at 11926 bytes on both.

The mistake: `git compare` reported *diverged, 6 ahead / 6 behind*, and I read "ahead" as "not merged." A **squash** merge leaves the branch's original commits as non-ancestors of main even though their content is in main, so ahead-by-N is exactly what a squash-merged branch looks like. Content, not ancestry, is the test.

No code was changed and nothing was pushed. #87 itself is still open despite its fix being in main.

Issue #115, opened 2026-09-08
Two markdown extraction paths read different frontmatter keys for ops:relatedTo, and a block-sequence `relatedTo:` emits nothing on either


ChristopherKahler:
### base version

`base 0.14.2` (tree at `610636e8`). Rust arm measured with the installed `base 0.14.1`; `src/extract/frontmatter.rs` is **byte-identical** at `v0.14.1` and `610636e8` (md5 `1f11911f13a2e7ada382e8f31770be8a`, zero commits touching it in between), so the measurement holds for this head. Python arm measured directly against the `610636e8` tree.

### Host and OS

Claude Code on WSL or Linux

### What happens

base has **two** markdown frontmatter extraction paths and they read **different keys** for the same `ops:relatedTo` predicate. No spelling works in both, and the spelling base's own published contract teaches produces zero edges on the primary path.

- `base sync` — Rust, `src/extract/frontmatter.rs:49` — matches the lowercased key `"relatedto"`.
- `base sync --ast` — Python, `scripts/ast/extractor.py:6127-6129` — reads `frontmatter.get("related")`, gated on `ontology: true`.
- The shipped contract `docs/markdown-ontology-protocol.md:35` documents **`related`** only, and all three of its worked examples (`:88`, `:102`, `:120`) write it as a YAML block sequence.

**Measured matrix.** Four documents, identical but for the frontmatter line, one `base sync` run in an isolated workspace, and four direct `extract_markdown()` calls:

| frontmatter | `base sync` (Rust) | `base sync --ast` (Python) |
|---|---|---|
| `related: [alpha-doc, beta-doc]` | **0** | 2 |
| `related:` + block list | **0** | 2 |
| `relatedTo: [alpha-doc, beta-doc]` | 2 | **0** |
| `relatedTo:` + block list | **0** | **0** |

Three consequences, in order of how quietly they bite:

1. **`relatedTo:` written as an ordinary YAML block sequence emits nothing on either path.** The Rust `parse_list` accepts only the inline `[...]` form; every fixture in `frontmatter.rs` (`:439`, `:601`) is inline, so nothing in the suite covers a block sequence. A user writing perfectly valid YAML gets silence from both extractors.
2. **The contract spelling is inert on the primary path.** A document written exactly as `docs/markdown-ontology-protocol.md` instructs contributes zero `ops:relatedTo` triples to `base sync`'s graph.
[...]

PR #154, opened 2026-09-09
fix(ast): read a PHP helper key in every plain quote style (#118)

### LINKED ISSUE #118: AST: the PHP helper-call branch reads only single-quoted arguments, so config("a.b") emits nothing while config('a.b') emits an edge
**Version:** `610636e8116c3c5f5e63d92f75397351bf51937c` (main, 2026-09-08)
**Host:** WSL2 Ubuntu 24.04 on Windows 10, python 3.12.3, `tree_sitter_php` from `pip install --user`

## What happens

`_extract_generic`'s helper-call branch (`scripts/ast/extractor.py:1935-1963`, the one whose own
comment reads `config('foo.bar') → uses_config edge to "foo"`) reads the helper's first argument by
walking for a child node of type `string`:

```python
for inner in arg.children:
    if inner.type == "string":
        for sc in inner.children:
            if sc.type == "string_content":
```

In `tree-sitter-php` a **single-quoted** literal is `string` and a **double-quoted** literal is
`encapsed_string`. Both carry a `string_content` child. So `config('app.name')` produces an edge and
`config("app.name")` produces nothing — from the same call, in the same file, differing only by
quote character.

Double quotes are the more common spelling in Laravel code, so the branch misses the majority case.

## Measured

One file, two adjacent statements, extracted with `extract_php` at `610636e`:

```php
<?php
class Helper { public function noop() { return 1; } }
class Svc {
    public function boot() {
        $a = config('Helper.name');    // single quotes
        $b = config("Helper.other");   // double quotes
        return [$a, $b];
    }
}
```

```
edges:
   contains  ...svc_php -> ...svc_helper
   method    ...svc_helper -> ...svc_helper_noop
   contains  ...svc_php -> ...svc_svc
   method    ...svc_svc -> ...svc_svc_boot
   uses_config ...svc_svc_boot -> ...svc_helper      <- ONE edge, from the single-quoted call
```

The grammar's own node types, dumped from the same parse, are the whole explanation:

```
function_call_expression b"config('Helper.name')"
  arguments -> argument -> string          b"'Helper.name'"  -> string_content b'Helper.name'
function_call_expression b'config("Helper.other")'
  arguments -> argument -> encapsed_string b'"Helper.other"' -> string_content b'Helper.other'
```

## Why it matters

The branch exists to record which config keys a class reads. Half of its real-world input is invisible
to it, silently — no warning, no counter. And because the relation it emits is `uses_config` (built by
f-string interpolation at `:1956` and spelled as a literal nowhere in the file), the edge it *does*
build is then dropped by the serializer under #107, so the defect has been double-masked: the common
spelling produced nothing, and the rare spelling produced an edge that never reached a map.

## Expected

Both quote styles are read. `encapsed_string` with a single `string_content` child is a constant
string exactly as `string` is; an interpolated double-quoted literal (`"{$x}.name"`) has more than one
child and should be skipped deliberately rather than by accident.

## Repro

```
python3 - <<'PY'
import sys; sys.path.insert(0, "scripts/ast")
from pathlib import Path
from extractor import extract_php
p = Path("/tmp/Svc.php"); p.write_text(
  "<?php\nclass Helper { public function noop() { return 1; } }\n"
  "class Svc { public function boot() { $a = config('Helper.a'); $b = config(\"Helper.b\"); } }\n")
for e in extract_php(p)["edges"]:
    print(e["relation"], e["source"], "->", e["target"])
PY
```

One `uses_config` edge, not two.

## Workaround

Write helper calls with single quotes. Not a reasonable ask of a codebase being mapped.

Found by `osprey` while building the #107 relation vocabulary — the site surfaced because
`uses_config` is the one relation name in `extractor.py` that no literal census can see.

ChristopherKahler:
Closes #118.

`config('a.b')` emitted a `uses_config` edge and `config("a.b")` emitted nothing, because the argument test was `inner.type == "string"` and tree-sitter types a double-quoted PHP literal `encapsed_string`.

## What this changes, and what it does not

**It makes the extractor read every plain quote style correctly and stop inventing targets. It does NOT by itself put config edges into a Laravel map** — see the measured finding at the bottom, which is ranked separately and not chased here.

## The filed claim named two node types. The installed grammar produces four

Measured against `tree-sitter-php` 0.24.1, not read off the grammar docs. `language_php` and `language_php_only` agree on all 18 spellings measured.

| spelling | node type | content lives in |
|---|---|---|
| `'a.b'` | `string` | `string_content` |
| `"a.b"` | `encapsed_string` | `string_content` |
| `<<<EOT` | **`heredoc`** | `value:` → `heredoc_body` → `string_content` |
| `<<<'EOT'` | **`nowdoc`** | `value:` → `nowdoc_body` → **`nowdoc_string`** |

Two things there are why a plausible fix still loses edges, and both are pinned by a test arm against the live grammar:

- **heredoc and nowdoc keep their text one level deeper**, so widening the type test while still reading the node's own children emits nothing for them, at exit 0.
- **a nowdoc's content node is `nowdoc_string`, not `string_content`**, so collecting only `string_content` reads an empty key for every nowdoc.

**Backticks stay out.** `` `a.b` `` is a `shell_command_expression` — shell execution, not a literal — and it carries a `string_content` child with **no** interpolation children, so a children-only test would admit it as a config key. The type set is a whitelist and the no-compile-time-value test is applied inside it.

## An argument with no compile-time value emits nothing and says so

One stderr line naming the file and the line. An interpolating literal has no single value, so any target emitted for it is invented, and a plausible-looking wrong target is worse than none because nothing will ever flag it — the same defect class as #115's 12-edges-per-character.

`escape_sequence` is treated the same way deliberately. The value exists, but decoding it needs a PHP escape decoder whose rules **differ between single and double quotes**, while both cheap readings fabricate: joining the siblings yields `ab` for `"a\tb"`.

**This deletes a live fabricated edge rather than declining a feature.** `config('a\'b.c')` was already admitted — type `string` — and the old break-after-first read stopped at the `escape_sequence` sibling and returned `a`. Its real key is `a'b.c`, whose first segment `a'b` resolves to nothing, so the extractor published an edge to `a` that the source never named. **Measured at the edge layer, not inferred:** the red run shows `esc_single_quote` emitting 1 edge to node `a` before this change.

## Acceptance: 13 named arms, both sides, three controls in every run

Base `4ce0e18e`. Red arm and green arm are told apart positively: the red arm requires the unfixed blob's sha256 **and** that `_php_literal_key` is absent; the green arm requires a different hash **and** that the symbol is present. Each run aborts rather than producing numbers if its arm is not the tree in front of it.

| arm | before | after | asserts |
|---|---|---|---|
| `plain_single` | 1 edge | 1 edge | unchanged (non-regression, and the standing emission control) |
| `plain_double` | **0** | **1 edge** → `a` | the filed defect |
| `plain_heredoc` | **0** | **1 edge** → `a` | not in the filed claim |
| `plain_nowdoc` | **0** | **1 edge** → `a` | not in the filed claim |
| `interp_var` | 0, no report | 0 **+ report** | `"a.$key"` |
| `interp_brace` | 0, no report | 0 **+ report** | `"a.{$key}"` |
| `interp_member` | 0, no report | 0 **+ report** | `"a.$o->k"` |
| `esc_single_quote` | **1 edge → `a`** | **0 + report** | the shipped fabricated edge, deleted |
| `esc_double_tab` | 0, no report | 0 **+ report** | `"a\tb.c"` |
| `esc_hex` | 0, no report | 0 **+ report** | `"a\x2eb"` |
| `backtick` | 0 | 0, **report absent** | shell execution stays out |
| `concat` | 0 | 0, **report absent** | `'a' . 'b'` |
| `non_helper` | 0 | 0, **report absent** | callee not in `helper_fn_names` |

**9 red before, 13/13 green after, zero mispredicted values.**

**Every arm asserts a 2-tuple — edge count AND report presence — because five of them read 0 edges on `main` for an unrelated reason:** the type test rejected `encapsed_string` wholesale, so nothing tested interpolation at all. An arm whose expected value is identical before and after proves nothing by itself. The three must-stay-silent arms assert report ABSENT, which catches the opposite defect: a fix that reports on everything it cannot resolve.

**Leg 5 (the exact key) is measured at the value layer**, because the edge target is the key's *first segment* only, so `a.b` and a truncation to `a` land on the same node and the edge layer cannot tell them apart. Calling `_php_literal_key` directly, all four plain spellings yield exactly `a.b` — no quotes, no escapes, no whitespace.

Controls in every run: **C1** per arm, three legs — the fixture parses with 0 `ERROR`/`MISSING` (checked directly against the grammar, since an ERROR-recovered tree still yields a function node), the caller node exists, and a `contains`/`method` edge exists. **C2** a helper call whose key resolves to nothing emits 0 edges and stays silent — a missing *target* is not an unknown *value*. **C3** five canary states, including that the report matcher rejects an incomplete line and rejects a conforming line bearing the wrong line number.

## Mutation matrix: 8 mutations, zero survivors

A green suite over a correct implementation proves nothing about the suite, so each plausible half-fix was applied to `extractor.py` and the suite re-run. Restore is verified by sha256 and a clean tree, not assumed.

| mutation | killed |
|---|---|
| `M1` two types only — the fix the filed claim implies | 3 legs |
| `M2` drop `nowdoc_string` | 3 legs |
| `M3` no body descent | 2 legs |
| `M4` join the escape siblings (option c) | 2 legs |
| `M5` report deleted | 2 legs |
| `M6` report on everything | 1 leg |
| `M7` revert the admission rule to `string` | 3 legs |
| `M8` report an unresolvable key | 1 leg |

7 of 9 legs are killed by something. The other two are unkillable by design and named as such: the canary asserts the matcher rather than the product, and the positive control fires when the *fixture* rots, not when the implementation does. `M8` was added after reading the first matrix — every mutation had died, but one leg asserting real product behaviour was reached by nothing, which is a hole in the matrix rather than in the suite.

## Tests

`scripts/ast/test_php_quote_styles.py`, 9 named legs, driven from `__main__` because CI runs these files as plain `python <file>` with no pytest. Reachability gate: 9 files, 63 tests, **0 unreachable**.

The first leg is a positive emission control asserted before anything counts a zero. Nine legs assert "zero edges", and all of them pass over a fixture where no edge could ever be emitted — which is not hypothetical: the first version of this fixture put the endpoint in a second file and **every leg read 0, including the single-quote baseline that was never broken**, because `label_to_nid` is built per file. A control proving the parser reached the file cannot tell that apart from a broken matcher; both read 0 at exit 0.

`scripts/ast/test_file_attribution.py` fails on Windows before this change and after it, with **byte-identical output at this commit and at its parent** — the known POSIX-separator assertion, not this change.

## Leg 8 on a real Laravel tree: +0, and the reason is not this fix

68 `.php` files, two arms over one file list, each in its own subprocess because Python caches modules. **Delta `+0` on both arms.** Split by layer rather than explained away:

| layer | measured |
|---|---|
| `config()` call sites | 50 |
| keys extracted correctly | **26** (`app.name`, `model.driver`, `model.api_key`, …) |
| argument not a literal | 24, every one `config([...])`, the setter form |
| **resolve in the same file** | **0** |
| **would resolve only across files** | **26** |
| no matching label anywhere | 0 |

Extraction works; the config branch never reaches the endpoint. A cross-file resolver already exists and PHP already feeds it — the ordinary call path appends to `raw_calls` under *"Callee not in this file — save for cross-file resolution in extract()"* — but **the config branch never appends on a miss**, so it drops the reference silently. Whether that global pass would accept a file-shaped target is unverified.

That finding is ranked separately in the work order and deliberately not chased here.

### DIFF
--- a/scripts/ast/extractor.py
+++ b/scripts/ast/extractor.py
@@ -1446,6 +1446,78 @@ _SCALA_CONFIG = LanguageConfig(
     import_handler=_import_scala,
 )
 
+# PHP produces FOUR node types for a plain string literal, and the content is not always
+# at the same depth. Measured against the installed grammar (tree-sitter-php 0.24.1)
+# rather than taken from the grammar's documentation:
+#
+#   'a.b'       -> string             string_content
+#   "a.b"       -> encapsed_string    string_content
+#   <<<EOT      -> heredoc            value: heredoc_body -> string_content
+#   <<<'EOT'    -> nowdoc             value: nowdoc_body  -> nowdoc_string
+#
+# `nowdoc_string` is why the body types have to be walked rather than assumed: a fix that
+# collects only `string_content` reads NOTHING for a nowdoc and emits no edge at exit 0.
+#
+# Backticks stay OUT. `` `a.b` `` is a `shell_command_expression` -- shell execution, not
+# a string literal -- and it carries a `string_content` child with no interpolation
+# children, so a children-only test would admit it as a config key. The type set is a
+# WHITELIST and the no-compile-time-value test is applied INSIDE it.
+_PHP_STRING_TYPES = frozenset({"string", "encapsed_string", "heredoc", "nowdoc"})
+_PHP_STRING_BODY_TYPES = frozenset({"heredoc_body", "nowdoc_body"})
+_PHP_STRING_CONTENT_TYPES = frozenset({"string_content", "nowdoc_string"})
+
+
+def _php_literal_key(node, source: bytes) -> "tuple[str | None, str | None]":
+    r"""The compile-time value of a PHP string literal, or the reason it has none.
+
+    Returns ``(key, None)`` for a plain literal; ``(None, reason)`` when the node IS a
+    string literal but has no single compile-time value; and ``(None, None)`` when the
+    node is not a string literal at all, so the caller keeps looking at later arguments.
+
+    The branch is on the node's CHILDREN, never on the quote character and never on the
+    node type alone: a double-quoted string with no interpolation is a plain literal that
+    tree-sitter still types ``encapsed_string``, which is the whole of #118.
+
+    ``escape_sequence`` counts as no-compile-time-value deliberately. The value does
+    exist, but decoding it needs a PHP escape decoder whose rules DIFFER between single
+    and double quotes, and both cheap readings invent data: joining the siblings yields
+    ``ab`` for ``"a\tb"``, and taking only the first sibling -- what this function
+    replaces -- yielded ``a`` for ``'a\'b.c'``, an edge to a target the source never
+    named.
+    """
+    if node.type not in _PHP_STRING_TYPES:
+        return None, None
+    body = node.child_by_field_name("value")
+    if body is None or body.type not in _PHP_STRING_BODY_TYPES:
+        body = node
+    parts: list[str] = []
+    for child in body.children:
+        if not child.is_named:
+            continue
+        if child.type in _PHP_STRING_CONTENT_TYPES:
+            parts.append(_read_text(child, source))
+        elif child.type == "escape_sequence":
+            return None, "escape sequence"
+        else:
+            return None, "interpolates"
+    return ("".join(parts) if parts else None), None
+
+
+def _report_no_literal_key(helper: str, reason: str, path_str: str, line: int) -> None:
+    """One stderr line per rejected helper argument, naming the file and the line.
+
+    Deliberately NOT widened to "report anything unresolvable". It fires only for an
+    argument to a recognised helper that IS a string literal without a compile-time
+    value. A key that resolves to no node in the file stays silent, because the key
+    itself was read correctly -- that is a missing target, not an unknown value.
+    """
+    print(
+        f"base-ast: {helper}() argument at {path_str}:{line} {reason} "
+        f"- no uses_{helper} edge emitted",
+        file=sys.stderr,
+    )
+
+
 _PHP_CONFIG = LanguageConfig(
     ts_module="tree_sitter_php",
     ts_language_fn="language_php",
@@ -2188,20 +2260,30 @@ def _extract_generic(
             if (callee_name and callee_name in config.helper_fn_names):
                 args_node = node.child_by_field_name("arguments")
                 first_key: str | None = None
+                # A literal that was recognised but has no compile-time value ENDS the
+                # search: it must not fall through to a later argument and emit an edge
+                # built from a default value.
+                no_literal_value = False
                 if args_node:
                     for arg in args_node.children:
                         if arg.type != "argument":
                             continue
                         for inner in arg.children:
-                            if inner.type == "string":
-                                for sc in inner.children:
-                                    if sc.type == "string_content":
-                                        first_key = _read_text(sc, source)
-                                        break
+                            if not inner.is_named:
+                                continue
+                            key, reason = _php_literal_key(inner, source)
+                            if reason is not None:
+                                _report_no_literal_key(
+                                    callee_name, reason, str_path,
+                                    node.start_point[0] + 1)
+                                no_literal_value = True
                                 break
-                        if first_key:
+                            if key is not None:
+                                first_key = key
+                                break
+                        if first_key or no_literal_value:
                             break
-                if first_key:
+                if first_key and not no_literal_value:
                     segment = first_key.split(".")[0]
                     tgt_nid = (label_to_nid.get(segment.lower())
                                or label_to_nid.get(f"{segment}.php".lower()))
--- a/scripts/ast/test_php_quote_styles.py
+++ b/scripts/ast/test_php_quote_styles.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+r"""A PHP helper call in ANY plain quote style produces its edge (#118).
+
+`config('a.b')` emitted a `uses_config` edge and `config("a.b")` emitted nothing, because
+the argument test was `inner.type == "string"` and tree-sitter types a double-quoted PHP
+literal `encapsed_string`. Double quotes are the more common Laravel spelling, so a real
+Laravel map was missing the majority of its config edges with nothing errored and nothing
+warned -- a smaller map that looks complete.
+
+THE FILED CLAIM NAMED TWO NODE TYPES. THE INSTALLED GRAMMAR PRODUCES FOUR
+-------------------------------------------------------------------------
+Measured against tree-sitter-php 0.24.1 rather than read off the grammar docs:
+
+    'a.b'      -> string             string_content
+    "a.b"      -> encapsed_string    string_content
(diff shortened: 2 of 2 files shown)


ChristopherKahler:
## Reviewer verdict: PASS — recorded on the PR because the grade happened out of band

All ten stamp conditions PASS. Record: `verification/base-0.14.3-raven/condor-preground-118.md`, sections 7 and 7.9. The pre-ground (T1-T10, Q1-Q8) was registered **before** any builder number was opened.

### The trap this PR walked into, and why the branch was updated

CI run `34357751016` was green — against `refs/pull/154/merge` = `7305ddea`, whose parents are `f46ea302` + **`4ce0e18e`**. That is a merge into a base that stopped existing at 13:45:36Z when #153 landed. `mergeable: true` also read green, from a computation timestamped 13:32:41Z. **A green flag can be older than the base it describes, and GitHub does not recompute it when main moves.**

Two of four jobs would have transferred to the real landing tree only by *argument*, at 1/5 residual risk. `update-branch` converted that to measurement.

| | before the update | after |
|---|---|---|
| head | `f46ea302` | `dc66155e` |
| base | `4ce0e18e` (stale) | `96ce58ca` (current main) |
| CI tested | a merge into the old base | the landing tree itself |
| T1 residual | 1/5, by argument | **0, by measurement** |

The landing tree is `6ff277be1a940b0ed0b5bbf3e9c51daddf9bd3fc`, computed independently two ways before the update and unchanged by it. **The merge condition was the tree, never the sha and never the flag.**

### What the fix does, and what it does not

Four plain quote styles now read: single, double, heredoc, nowdoc. Interpolated arguments emit **nothing, deliberately**, and say so on stderr: `config() argument at ...:19 interpolates - no uses_config edge emitted`. Skipped **and** reported, on the channel an operator reads.

Measured end to end: `uses_config` edges **1 → 4**, `usesConfig` in the emitted TTL **2 → 8**.

**It also deletes a fabricated edge.** `config('a\'b')` was admitted before this change and emitted an edge to target `a` — a target the source never named.

**What it does not do:** on a real Laravel layout the config key names a file in `config/` while the call sits in a controller elsewhere, and cross-file resolution for that branch never fires. Measured on a 68-file tree: same-file `1 → 2`, cross-file `0 → 0`. That is a separate ranked item, not this PR's problem.

### Verification

- **9 named test arms**, `ran 9 of 9 legs rc=0`, with the reachability gate ahead of them reading `defined=9 unreachable=0`.
- **7 mutation arms, 6 detected, 6 distinct reasons.** The strongest: admitting interpolated keys produces an edge to a target the source never named, and the test says exactly that.
- The defective `if inner.type == "string"` occurs **1 time** at `96ce58ca` and **0** at the head — a genuine removal, not a bare zero, with a positive control finding 448 defs and a canary reading 0.
- Windows baseline measured both sides: no new failures, none fixed, the single pre-existing red identical in name, location and reason.
- Final CI `34363041681` on the landing tree: 4 jobs, **33 steps, 0 non-success, 0 skipped**.

PR #156, opened 2026-09-09
fix(ast): a PHP config key resolves to its config file in another file

ChristopherKahler:
Closes rank 05, `N-HELPER-NEVER-RESOLVES` (`BASE-WORK-ORDER.md:1033`). Branched off `main 96df6829`. **DO NOT MERGE** — `condor` grades, `auk` merges (law 6).

## The finding

The config branch resolved its key against `label_to_nid`, which holds only the **current file's** nodes. `config/app.php` is a different file, so `config('app.name')` in a controller could never resolve, and the reference was dropped with **no edge and no report**.

Measured on `Documents/caddy-baseline` (68 `.php`) at `96df6829`: **26 keys read, 0 edges emitted, and 0 of the 26 resolved in-file.** The feature was silent on a real Laravel app.

## The blocker this PR is shaped around

Handing the miss to the cross-file promotion in `extract()` is the obvious fix and is wrong on its own: that promotion **hard-coded** `"relation": "calls"`. Routed in unchanged it asserts that a controller **calls** `config/app.php` — fabricated relationship data of the same class #115 and #118 shipped fixes for, and **worse than the silence it replaces, because a wrong edge is believed.**

So `raw_calls` entries gained an **optional** `relation`, read as `rc.get("relation", "calls")`. The key is never required: `raw_calls` is **persisted to the per-file cache**, so a cache written before this change is read after it. The five other producers are untouched and keep emitting `calls` by default.

A colliding file-shaped target is **reported**, not dropped — measured on the same tree: 66 distinct basenames across 68 files, `auth.php` twice, and one of the two is among the 11 `config/` files, so a Laravel app with `config/auth.php` and any other `auth.php` lost that key with no error. Preferring a `config/` directory match was **rejected**: framework-specific logic inside a pass built for every language is a defect waiting for its second framework.

## Acceptance

Instrument `accept05.py`, sha256 `7d42135174a0ab4c1b5ff5e378dca47d11c0c9ba2fea802a10efafc59a638774` — **the same build produced both records below.** Two-arm provenance gate, told apart positively: red requires the `96df6829` blob hash **and** the new symbol absent; green requires a different hash **and** the symbol present. Each aborts rather than producing numbers.

| # | arm | RED (`96df6829`) | GREEN (`5d9b0a32`) | what it discriminates |
|---|---|---|---|---|
| 1 | `cross_file_resolves` | **FAIL** 0 edges | PASS 1 edge → `a.php` | the lane never reaching the resolver |
| 2 | `relation_is_uses_config` | **FAIL** no relation | PASS `uses_config` | ⛔ **the blocker.** Arm 1 passes for a fix that omits the relation |
| 3 | `in_file_still_resolves` | PASS 1 edge | PASS 1 edge | a change that moves resolution to the global pass, breaking #118 |
| 4 | `no_endpoint_anywhere` | PASS 0, silent | PASS 0, silent | a fix that reports every non-emitting call |
| 5 | `collision_reports` | **FAIL** 0 reports | PASS 0 edges + 1 report | a colliding target dropped silently |
| 6 | `stale_cache_defaults` | **FAIL** | PASS defaults to `calls`, no raise | a required key (raises) vs a wrong default (mis-relates) |
| 7 | `ordinary_calls_unchanged` | PASS 1 `calls` | PASS 1 `calls` | the default path; **also C1** |
| 8 | `interpolating_never_reaches_raw_calls` | PASS | PASS 0 edges, no leak, 1 report | a non-literal reaching cross-file resolution |
| 9 | `member_call_still_skipped` | PASS 0 | PASS 0 | the `is_member_call` guard |
| 10 | `real_tree_delta` | **FAIL** 0 edges | PASS 14 edges | the finding on the real corpus |

**RED: 5 of 5 must-fail arms failed, all four controls green.** GREEN: **10 of 10 arms, all four controls green, rc 0.**

Controls in every run, void by name with a nonzero rc: **C1** the promotion can fire at all, asserted **first** (#118 lost thirteen arms to a fixture whose target was never resolvable); **C2** a key that reads perfectly and matches nothing stays silent; **C3** five canary red states including a report matcher that rejects the wrong line number and the wrong target; **C4** every fixture parsed directly against the grammar, 0 `ERROR` and 0 `MISSING`.

### Real-tree delta, accounted key by key (`delta05.py`, rc 0)

```
keys read                : 26   = 26 handed to cross-file + 0 resolved in-file
  emitted cross-file     : 14   (model.php 13, app.php 1)
  emitted in-file (#118) :  0
  = uses_config edges    : 14   (run: 14)
  reported ambiguous     :  0   (run: 0)
  correctly silent       : 12   (duplicate caller/target pairs deduped by existing_pairs)
  UNACCOUNTED            :  0
```

Every key read is in exactly one bucket and the buckets predict the run's edge and report counts exactly. **0 ambiguous reports is the predicted value**, not a miss: MEASURE-05 recorded that this corpus never calls `config('auth.*')`, so the collision path has no input here. Arm 5 is what proves that path fires.

## Committed legs

`scripts/ast/test_php_config_cross_file.py` — 10 legs, driven from `__main__` because `ci.yml` runs each file as plain `python <file>` with no pytest. Reachability gate: `shape=A defined=10 unreachable=0`, run total `files=10 defined_tests=73 unreachable=0`.

The positive control runs **first and by name**: if the promotion cannot emit an ordinary cross-file `calls` edge, the file prints `VOID` and exits 2 rather than going green on vacuous zeros. **Against unfixed `96df6829`, four legs fail while that control passes throughout.**

## Mutation matrix (`mutate05.py`, rc 0)

Nine plausible half-fixes. **0 survivors, 0 unapplied, tree clean and `extractor.py` sha256 restored after every revert.**

| mutation | detected by |
|---|---|
| `M1_relation_not_carried` | `collision_reports`, `cross_file_resolves`, `relation_is_uses_config`, `stale_cache_defaults` |
| `M2_relation_required` | 9 of 10 legs — **blunt**, see below |
| `M3_wrong_default` | `ordinary_collision_stays_silent`, `ordinary_cross_file_call_still_resolves`, `stale_cache_defaults` |
| `M4_collision_silent` | `collision_reports` |
| `M5_never_appends` | `collision_reports`, `cross_file_resolves`, `relation_is_uses_config`, `stale_cache_defaults` |
| `M6_bare_segment_target` | `collision_reports`, `cross_file_resolves`, `relation_is_uses_config`, `stale_cache_defaults` |
| `M7_member_guard_removed` | `member_call_still_skipped` |
| `M8_report_widened_to_all` | `ordinary_collision_stays_silent` |
| `M9_context_kept_on_uses` | `relation_is_uses_config` |

**Two findings from the matrix itself, reported rather than smoothed over.**

**(a) `M2` and `M3` exit 2 through the positive control, so the later legs never ran.** That is the control doing its job (law 48: a failed control voids the run and must not print counts) and it **masks** every subsequent verdict. "Killed by the control" and "killed by the leg that discriminates it" are different claims, and reporting the first as the second credits a leg that was never reached. So each leg is re-run in **isolation**, one subprocess per leg, to answer what the masked run cannot.

**(b) After isolation, `M2` kills 9 of 10 legs** — it raises a `KeyError` inside `extract()`, so every leg that extracts anything dies. Crediting those legs as proven is the **inverted** form of the error law 49 names: it reports a leg as discriminating a property when all it did was witness a global crash. A mutation killing more than half the suite is therefore graded **blunt** and credits no leg with anything specific.

Result: **7 of 10 legs proven by a targeted mutation. 0 killed by nothing. 3 killed only by the blunt `M2`** — `in_file_resolution_unchanged`, `interpolated_key_never_reaches_cross_file`, `missing_target_stays_silent` — and all three carry a structural reason for being unreachable (they guard #118's lane and the `candidates == 0` path, neither of which any mutation here edits). **Real matrix gap: NONE.**

`test_ordinary_collision_stays_silent` was added **while writing the matrix, not after it ran**: without it, dropping the report's `relation != "calls"` condition changes real behaviour on any large repo and no leg notices.

## Two spec defects found by standing order 6, corrected in place

Both were wrong **constants**, not unclear purposes. `auk` accepted both.

1. **The provenance hash was from the wrong tree.** SPEC-05 pinned `8384f0d8…5496b9` and called it "the `96df6829` blob". It is `.cache/marten-127/main-4ce0e18e/scripts/ast/extractor.py` — an **8552-line** file, in a tree whose HEAD is `96ce58ca`, which does not contain commit `96df6829` at all. It was copied forward from `accept118.py`, where it is correctly pinned to `MAIN_SHA = 4ce0e18e`. The real `96df6829` blob is sha256 `3326c286…48e049` (git blob `3fb965c7`, 8634 lines, 370480 bytes, LF — both LF and CRLF forms hashed, neither matches). **Pinned as written the red arm can never match, so the gate could only abort, which looks like protection and is not.**
2. **`if len(candidates) != 1:` is at `:8453`, not `:8452`.** Every other anchor verified exact by construct text, including both producer lists.

## Pre-existing, not from this PR

`scripts/ast/test_file_attribution.py` fails **identically on the pristine `96df6829` worktree** on this Windows box (`src/alpha/mod.rs has no entities attributed to it`). Same failure with and without this change, so it is not a regression here; CI on Linux is the reading that matters. Flagged, not touched.

## Evidence

`~/.base-gbl/verification/base-0.14.3-heron/`: `SPEC-05-acceptance.md` (with the corrections section), `MEASURE-05.md`, `accept05.py`, `ACCEPT-05-red.{txt,json}`, `ACCEPT-05-green.{txt,json}`, `delta05.py`, `mutate05.py`, `MUTATE-05.{txt,json}`.

### DIFF
--- a/scripts/ast/extractor.py
+++ b/scripts/ast/extractor.py
@@ -1518,6 +1518,38 @@ def _report_no_literal_key(helper: str, reason: str, path_str: str, line: int) -
     )
 
 
+def _report_ambiguous_cross_file_target(
+    relation: str, callee: str, candidates: list[str], path_str: str, location: str
+) -> None:
+    """One stderr line when a helper reference names a target that is not unique.
+
+    Scoped to raw_calls entries carrying their own `relation` -- the helper-reference
+    lane -- for the same reason _report_no_literal_key is not widened. An ordinary
+    cross-file call to a common short name (log, execute, find) collides by design and
+    its silence is the DOCUMENTED behaviour of the uniqueness gate, not a defect; a
+    helper reference whose key read perfectly and named exactly one file that happens
+    to share a basename is a silent loss, which is the shape #118 shipped a fix for.
+
+    Measured on a real Laravel tree before scoping this: 897 raw_calls entries, and
+    entries hitting len(candidates) > 1 in the ordinary-call lane numbered ZERO -- so
+    the narrow scope is not about volume today, it is about the collision case the
+    comment on the uniqueness gate itself names for repos larger than that one.
+
+    REPORTS rather than preferring a candidate. A directory-based tie-break (prefer a
+    `config/` match) would put framework-specific logic inside a pass built for every
+    language, which is a defect waiting for its second framework.
+    """
+    # raw_calls stores source_location as "L<n>"; the report says ":<n>" so it reads the
+    # same as _report_no_literal_key above. One report convention, not two.
+    line = location[1:] if location[:1] == "L" else location
+    print(
+        f"base-ast: {relation} target {callee!r} at {path_str}:{line} matches "
+        f"{len(candidates)} nodes ({', '.join(sorted(candidates))}) "
+        f"- ambiguous, no {relation} edge emitted",
+        file=sys.stderr,
+    )
+
+
 _PHP_CONFIG = LanguageConfig(
     ts_module="tree_sitter_php",
     ts_language_fn="language_php",
@@ -2303,6 +2335,28 @@ def _extract_generic(
                                 "source_location": f"L{line}",
                                 "weight": 1.0,
                             })
+                    elif not tgt_nid:
+                        # The key read fine but names no node in THIS file. Hand it to
+                        # the cross-file pass in extract() the way an unresolved call is
+                        # handed over -- but carrying its OWN relation, because the
+                        # promotion hard-codes "calls" and a config key is a reference,
+                        # not a call. Routing this in without the relation would assert
+                        # that a controller CALLS config/app.php: fabricated
+                        # relationship data, and worse than the silence it replaces
+                        # because a wrong edge is believed.
+                        #
+                        # The target is the ".php" form. Measured on a real Laravel
+                        # tree: the bare segment is not a key in the global label index
+                        # at all (0 hits) and only "<segment>.php" resolves -- and it is
+                        # the form this branch already computes on the line above.
+                        raw_calls.append({
+                            "caller_nid": caller_nid,
+                            "callee": f"{segment}.php",
+                            "relation": f"uses_{callee_name}",
+                            "is_member_call": is_member_call,
+                            "source_file": str_path,
+                            "source_location": f"L{node.start_point[0] + 1}",
+                        })
 
             # Service container bindings: $this->app->bind(Foo::class, Bar::class)
             if (node.type == "member_call_expression"
@@ -8447,10 +8501,26 @@ def extract(
         if rc.get("is_member_call"):
             continue
         candidates = global_label_to_nids.get(callee.lower(), [])
+        # An entry may carry its own relation (a helper reference such as
+        # config('app.name') is a REFERENCE to a file, not a call to it). The key is
+        # OPTIONAL and read with a default: raw_calls is persisted to the per-file
+        # cache, so a cache written before this field existed is read back after it
+        # and must default rather than raise.
+        relation = rc.get("relation", "calls")
         # Skip ambiguous names that resolve to multiple nodes — these are
         # common short names (log, execute, find) with no import evidence
         # to pick the right target; emitting all edges inflates god_nodes.
         if len(candidates) != 1:
+            # A helper reference that collides is a SILENT LOSS: its key read
+            # perfectly and named one file, which happens to share a basename with
+            # another. Report it rather than preferring a candidate. Ordinary calls
+            # keep the documented silence — colliding on a common short name is the
+            # behaviour this gate exists for, not a defect.
+            if relation != "calls" and len(candidates) > 1:
+                _report_ambiguous_cross_file_target(
+                    relation, callee, candidates,
+                    rc.get("source_file", ""), rc.get("source_location", ""),
+                )
             continue
         tgt = candidates[0]
         caller = rc["caller_nid"]
@@ -8473,17 +8543,22 @@ def extract(
             else:
                 confidence = "INFERRED"
                 confidence_score = 0.8
-            all_edges.append({
+            edge = {
                 "source": caller,
                 "target": tgt,
-                "relation": "calls",
-                "context": "call",
+                "relation": relation,
                 "confidence": confidence,
                 "confidence_score": confidence_score,
                 "source_file": rc.get("source_file", ""),
                 "source_location": rc.get("source_location"),
                 "weight": 1.0,
-            })
+            }
+            # "context": "call" describes a call site. A helper reference is not one,
+            # and the in-file uses_* emission at the config branch does not set it
+            # either — so the two lanes agree on the shape they produce.
+            if relation == "calls":
+                edge["context"] = "call"
+            all_edges.append(edge)
 
     # Relativize source_file fields so paths are portable across machines (#555)
     for item in all_nodes + all_edges:
--- a/scripts/ast/relation_vocabulary.py
+++ b/scripts/ast/relation_vocabulary.py
@@ -31,6 +31,23 @@ assumptions (law 31 — a guard is only proven against the codebase):
      this single site. Resolving it means reading the guard, so that widening
      `helper_fn_names` widens the vocabulary and fails this test until the
      table is updated, rather than silently dropping a new relation.
+  7. the SAME f-string, written INLINE in the slot rather than bound to a name
+     first. Shape 6 only ever looked for an `ast.Assign`, so an author who put
+     `"relation": f"uses_{callee_name}"` straight into a `raw_calls` dict — the
+     obvious spelling, and identical in meaning — landed in `unresolved`. The
+     reduction is the same closed-set walk; only where the JoinedStr sits
+     differs, so shapes 6 and 7 share one resolver.
+  8. a RELAY: a name bound to `<mapping>.get("relation", <const>)`, which hands
+     on a relation some OTHER site in this file already wrote. It resolves to
+     the default, and its non-default range needs no separate resolution
+     because shape 2 already visits every site in this file that writes a
+     `"relation"` key — a relay cannot introduce a relation no writer spelled,
+     and if a writer's own slot is unresolvable then THAT site fails the census.
+     The closure is only valid while every write goes through a dict literal, so
+     the resolver first proves there is no `<expr>["relation"] = ...` subscript
+     write anywhere in the file; one such write and the relay goes back to
+     `unresolved` rather than being trusted. That is what keeps this from being
+     a narrowing: the shape is falsifiable by a single line of new code.
 
 Anything the resolver cannot reduce to constants is returned in `unresolved`, and
 the caller is expected to FAIL on a non-empty list rather than report a smaller
@@ -357,6 +374,10 @@ def extractor_relations(source_path: Path | str) -> Vocabulary:
         if isinstance(n, (ast.FunctionDef, ast.AsyncFunctionDef)) and n.name in EDGE_FNS
     )
     counts["forwarders"] = len(forwarders)
+    # Shape 8's precondition, computed ONCE and reported so a reader can see it was
+    # checked rather than assumed. A non-empty list revokes every relay.
+    subscript_writes = _subscript_writes_key(tree, "relation")
+    counts["subscript_writes_relation_key"] = len(subscript_writes)
 
     def record(shape: str, node: ast.AST, slot: ast.AST | None) -> None:
         if inside_helper(getattr(node, "lineno", 0)):
@@ -378,6 +399,22 @@ def extractor_relations(source_path: Path | str) -> Vocabulary:
             if expanded:
                 vocab.relations.update(expanded)
                 counts["via_closed_set"] = counts.get("via_closed_set", 0) + 1
+                return
+        # shape 7: the same f-string written INLINE in the slot
+        if isinstance(slot, ast.JoinedStr) and scope is not None:
+            inline = _expand_joinedstr(slot, node, scope, closed_sets)
+            if inline:
+                vocab.relations.update(inline)
+                counts["via_inline_fstring"] = counts.get("via_inline_fstring", 0) + 1
+                return
+        # shape 8: a relay of a relation another site in this file wrote. Valid only
+        # while no subscript write can smuggle in a value no dict literal spells.
+        if (isinstance(slot, ast.Name) and scope is not None
+                and not subscript_writes):
+            relayed = _relay_bindings(scope, slot.id, "relation")
+            if relayed:
+                vocab.relations.update(relayed)
+                counts["via_relay"] = counts.get("via_relay", 0) + 1
                 return
         if isinstance(slot, ast.Attribute):
             per_class = _dataclass_field_constants(tree, slot.attr)
@@ -406,17 +443,56 @@ def extractor_relations(source_path: Path | str) -> Vocabulary:
     return vocab
 
 
+def _expand_joinedstr(
+    joined: ast.JoinedStr,
+    at: ast.AST,
+    scope: "_Scope",
+    closed_sets: dict[str, set[str]],
+) -> set[str]:
+    """The shared half of shapes 6 and 7: reduce ONE f-string to constants using the
+    guards that dominate the position `at`.
+
+    Split out of `_expand_fstring_binding` when shape 7 arrived. The reduction never
+    depended on the JoinedStr being the right-hand side of an assignment — only on
+    which guards enclose it — so a bound f-string and an inline one are the same
+    problem read at two different nodes.
+
+    Returns the full cross-product, or an EMPTY SET when any part of the chain is
+    open. An empty return means "unresolved", never "no relations".
+    """
+    pieces: list[set[str]] = []
+    for part in joined.values:
+        literal = _const_str(part)
+        if literal is not None:
+            pieces.append({literal})
+            continue
+        if not (isinstance(part, ast.FormattedValue)
+                and isinstance(part.value, ast.Name)):
+            return set()
+        guards = _dominating_guards(scope.node, at)
+        fields = guards.get(part.value.id) or set()
+        if len(fields) != 1:
+            # No dominating guard, or more than one — the name is not provably
+            # closed at this site, so the whole slot is unresolved rather than
+            # partially guessed.
+            return set()
+        members = closed_sets.get(next(iter(fields)))
+        if not members:
+            return set()
+        pieces.append(set(members))
+    combos = {""}
+    for piece in pieces:
+        combos = {prefix + suffix for prefix in combos for suffix in piece}
+    return combos
+
+
 def _expand_fstring_binding(
     name: str,
     scope: "_Scope",
     closed_sets: dict[str, set[str]],
 ) -> set[str]:
     """Shape 6: a name bound to an f-string over other names the same scope has
-    constrained to a closed frozenset.
-
-    Returns the full cross-product, or an empty set when any part of the chain
-    is open — an empty return means "unresolved", never "no relations".
-    """
+    constrained to a closed frozenset."""
     out: set[str] = set()
     for node in ast.walk(scope.node):
         if not (isinstance(node, ast.Assign) and len(node.targets) == 1):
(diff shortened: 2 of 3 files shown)


ChristopherKahler:
## CI on `5d9b0a32` failed on one test, and the gate was right. Fixed in `638fe2e5`.

**DO NOT MERGE.** A green CI here is not a go: `condor` grades, `auk` merges, and a **merge freeze on `main`** is in force.

Run [`34367571959`](https://github.com/ChristopherKahler/base/actions/runs/34367571959), read at **step** level: `python-ast-tests`, `clippy` and `release-invocations` green; **`test / cargo test` failed on exactly one test**, `every_emittable_relation_has_a_predicate` (`tests/ast_extension_registry.rs:294`).

`relation_vocabulary.py` reduces every relation slot in `extractor.py` to constants so no emittable relation can escape the TTL predicate table. It reported:

```
emittable 27   mapped 27   missing 0 []   dead 0 []
unresolved 2 ["extractor.py:8546  Name(id='relation', ctx=Load())",
              "extractor.py:2352  JoinedStr(values=[Constant(value='uses_'), FormattedValue(..."]
```

**Nothing was unmapped.** It failed because two slots this PR added would not reduce — and law 23 is that a resolver which silently skipped sites proves less than it claims, so an unresolved site is a failure of the test, never a smaller vocabulary. The test's own message: *teach the census the new shape, do not narrow the claim.*

### Two shapes, enumerated in the module docstring beside the six already there

**Shape 7 — an inline f-string.** Shape 6 resolved `relation = f"uses_{callee_name}"` by searching for an `ast.Assign`, so the same f-string written straight into a dict slot fell through. The reduction never depended on the assignment, only on **which guards dominate the position** — so 6 and 7 now share one resolver read at two different nodes.

**Shape 8 — a relay.** `rc.get("relation", "calls")` hands on a relation another site in this file already wrote. It resolves to the **default**; its non-default range needs no separate resolution because **shape 2 already visits every dict literal in this file carrying a `"relation"` key**. A relay cannot introduce a relation no writer spelled, and an unresolvable writer fails at its own site.

That closure holds **only while every write goes through a dict literal**, so the resolver first proves there is no `<expr>["relation"] = ...` subscript write in the file and **publishes the count it found** (`path subscript_writes_relation_key 0`). One such write revokes every relay.

### Why shape 8 is not a narrowing — proved, not argued

A shape added so a gate can pass is a narrowing unless it can still go red, and a shape never seen red is decoration that happens to print PASS. Both new shapes were run against fixtures built to defeat them (`verification/base-0.14.3-heron/vocab_shapes_red_state.py`, rc 0):

| arm | expects | result |
|---|---|---|
| `A1_guarded_inline_fstring_resolves` | resolves → `uses_config` | PASS |
| `A2_unguarded_inline_fstring_is_unresolved` | **unresolved ≥ 1** | PASS |
| `A3_relay_with_literal_writers_resolves` | resolves → `calls` | PASS |
| `A4_relay_with_subscript_write_is_unresolved` | **unresolved ≥ 1** | PASS |

A2 and A4 are the arms that matter. **A1 and A3 are their controls** — without them a resolver that reduced *nothing* would pass both red arms and look rigorous.

### Census after

```
unresolved 0 []          via_inline_fstring 1     via_relay 1
subscript_writes_relation_key 0                    slots visited 219
emittable 27   mapped 27   missing 0 []   dead 0 []
```

Both LINE-EXACT pins the test asserts still hold: `path forwarders 1`, `path via_closed_set 1`. The `FLOOR not CLOSURE` line still prints, now over 12 enumerated shapes instead of 9.

`cargo` is not on PATH on the build box, so the Rust assertions were checked by running the test's own embedded Python verbatim from `scripts/ast/` as cwd. CI is the only real reading.

### The acceptance record is unaffected

`638fe2e5` touches **only** `relation_vocabulary.py`. `extractor.py` is byte-identical (sha256 `15bea3aecf8d6545aa85667b58cb3832e3a0d0b2a0c621728b93aa0f989ea836`) to the tree the GREEN record in the PR body was taken on, so that record stands exactly as written.

### One more thing worth stating plainly

**Zero of the 26 config keys on the real tree resolved in-file.** #118's in-file lane emits nothing at all on a real Laravel app — every one of those 26 keys needed this cross-file lane. That is what a user actually gains here.


---

**CI on `638fe2e5`: run [`34368412253`](https://github.com/ChristopherKahler/base/actions/runs/34368412253) is GREEN — 4 jobs, 33 steps, read at STEP level (law 18), every step `success` or `skipped`.** Still **DO NOT MERGE**.

ChristopherKahler:
**DO NOT MERGE — reviewer grades, `auk` merges.**

Measurement record for the rebase onto `main`, re-taken in full on the **rebased** head. Every
number below was taken after the rebase; nothing is carried forward from the pre-rebase head.

| | |
|---|---|
| rebased head | `13a6de718fbd6e6f203148cb93ba4c5a26e08f15` |
| base | `main` = `4bc5c1cfbc1cef919cc4474f9148f6e6b4aa5fd0` |
| rebase conflicts | **0** |
| compare `main...head` | `ahead`, ahead_by 3, **behind_by 0** |
| files | 3, all under `scripts/ast/`, **0 `.rs`**, +575 −32 |
| corpus | `caddy-baseline` frozen by `git archive HEAD`, identity `5041c93887bf770097539b5e64238cac1490f852`, 242 files |

## The finding this fixes

On a real Laravel tree the extractor read every `config()` key correctly and resolved **none** of
them, because the config-helper branch looked its key up only among the *current file's* nodes,
while every config key names a file in `config/` called from a controller elsewhere. The shipped
map carried **zero** `uses_config` edges on a real app — a shipped claim that was false.

**Measured here, and worse than "zero edges": `main` says nothing at all about those keys.** Its
stderr over the whole corpus is 7 lines and not one concerns `config()`. A control proves that same
binary's report channel is live (below), so this is silence, not absence of a channel. A user got no
edge **and** no diagnostic.

## Before and after, same frozen corpus, same 242-file list, one `python3` process per arm

| reading | before (`main 4bc5c1c`) | after (`head 13a6de71`) |
|---|---|---|
| **`uses_config` edges** | **0** | **13** |
| `usesConfig` occurrences in the emitted TTL | 0 | 26 |
| ambiguity reports on stderr | 0 | **1** |
| distinct predicates in the TTL | 7 | 8 |
| total edges in the map | 2,355 | 2,368 |

### The population, named — because a count is not a population

Every reference classified individually; no aggregate is derived by subtracting another:

> The extractor read **26** config-helper references across **9** PHP files.
> **13** became `uses_config` edges.
> **1** was **reported ambiguous** and deliberately not emitted.
> **12** collapsed onto an edge already emitted for the same source and target.
> **Nothing is unaccounted for.** `13 + 1 + 12 = 26`.

The 26 is the extractor's own count, read from `extract_php()`'s `raw_calls` where it is produced.
The classification was cross-checked against the edge list counted a different way — 13 and 13,
agree — and the reconciler refuses to present a population if the two disagree. All 13 edges point
at `config/model.php`.

### Why one detector reads 26 and the other reads 13, measured rather than reported as a conflict

Two detectors on two channels: the emitted TTL (what a consumer of `base sync --ast` reads) and the
in-process result. They read 26 and 13. That is **not** a discrepancy:

| shape | count |
|---|---|
| TTL lines carrying the token `usesConfig` | 26 |
| …as the predicate of a plain triple | 13 |
| …inside an `<< >>` annotation carrying the edge's confidence | 13 |
| …fitting neither shape | **0** |

Each edge is written to the TTL twice, so `26 = 13 × 2`. One detector counts token occurrences and
the other counts edges — two correct counts of two different populations. The reconciler **refuses**
if any line fits neither shape, so that sum is a reading and not arithmetic.

**Scope clause, stated rather than oversold:** the two detectors are independent across the
*serialization* axis; they are **not** independent across the *extraction* axis, where a single
break moves both.

## Acceptance legs

| # | leg | reading |
|---|---|---|
| 1 | `uses_config` edges before | **0**, behind a positive control (7 predicates, 4,712 occurrences in the same TTL) and a must-fail canary |
| 2 | `uses_config` edges after, population named | **13** of 26 references; 1 reported, 12 collapsed |
| 3 | `test_php_config_cross_file.py` rc | **0** (absent on `main`) |
| 4 | every other `scripts/ast/test_*.py` rc | **10 of 10 pass on head, 9 of 9 on main**, per-file rc, files-visited printed |
| 5 | rebase conflict count | **0** |
| 6 | CI at STEP level on the rebased head | see below |
| 7 | the three files' blob shas across the rebase | **identical** — `5749402c…`, `ffda1c9a…`, `b7b0c194…` |
| 8 | `cargo test` | **not required for this lane** — zero `.rs` files change. Stated rather than skipped silently; CI runs it and leg 6 reads it at step level |
| 9 | the ambiguity report names `app.php` | **PASS**, verbatim below; the same grep on `main` reads **0**, which is the leg's own control |

### Leg 9, verbatim

```
base-ast: uses_config target 'app.php' at .../app/Http/Middleware/HandleInertiaRequests.php:42 matches 2 nodes (bootstrap_app_php, config_app_php) - ambiguous, no uses_config edge emitted
```

Leg 9 exists because an edge count cannot separate a key that **resolved** from one that was
**silently dropped**. `config('app.name')` is expected to be reported rather than resolved, because
`bootstrap/app.php` and `config/app.php` share a basename. Had the fix regressed into dropping it
silently, both detectors would still have read `> 0` from the `model.*` keys and passed.

**Leg 9 also survives a WARM cache** — measured, not assumed. Two consecutive runs over the same
frozen corpus with nothing purged: cold emits the report, warm emits it too, and the TTL is
byte-identical. The warm run genuinely reused cache (242 uncached on run 1, 157 on run 2). The
guarantee is not cold-run-only.

### The must-fail canary behind the "before = 0"

A zero cannot distinguish *the thing is not there* from *I cannot see*. Run on the **same** `main`
extractor that produced the zero, on a fixture chosen to be independent of the fix under test:

| capability | reading |
|---|---|
| it CAN emit a `uses_config` edge (in-file fixture `main` already ships) | **1 edge** |
| that edge CAN reach the TTL | **2** occurrences, `code:… ops:usesConfig code:… .` |
| its stderr report channel CAN carry a line | **1** line, `config() argument … interpolates - no uses_config edge emitted` |
| a non-helper call does **not** match | **0** edges |

## Reconciling with the original build's 14

The earlier build recorded `26 = 14 edges + 12 deduped`. This run reads `26 = 13 edges + 1 reported
+ 12 deduped`. **The 12 is identical and the population is identical**; exactly one reference moved
from the edge bucket to the report bucket — `app.name`, because `bootstrap/app.php` now exists in
the corpus and makes that basename ambiguous. The mechanism is measured here; the claim about what
the earlier corpus contained is an **inference**, since that corpus state no longer exists.

## Two instrument defects found in my own harness, disclosed

Neither is a defect in this PR. Both are recorded because an instrument that misled once and was
corrected is worth more than one that never reported its own failures.

1. **A custody check that tested for PRESENCE when only a CHANGE could be a leak.** It reported a
   cache directory in the client tree as a failure; the directory pre-dates this work by six days.
   Settled three ways, one needing no clock at all: `collect_files()` creates 0 cache dirs on a
   clean root while a real `extract()` creates 1, and `collect_files()` is the only call any script
   of mine made against that tree. The corrected assertion reads **unchanged** by file count and
   newest mtime, and passed on both arms.
   *Also caught here:* my first clock probe returned "nothing touched" and I refused the reading,
   because **its own positive control also read 0** — `find -newermt` parses a bare timestamp as
   local time and I had handed it a UTC figure, putting the cutoff in the future.

2. **An absolute `BASE_AST_OUT` breaks `scripts/ast` test isolation.** `cache.py:_cache_base()`
   returns an absolute value verbatim, ignoring the `root` it is handed, so every fixture in every
   test shares one cache. With it set, three tests go red — one of them reading `the rejection was
   silent`, squarely in this lane's subject matter. The control that separates harness from branch:
   **`main`, which carries none of this branch's code, fails identically.** Unset, head is 10/10 and
   main 9/9. This is pre-existing on `main` and is filed as an unranked finding, not fixed here.
   The same family bit my canary a second time: an unpurged cache replayed a fixture, and **a replay
   reproduces edges but not stderr**, so one control silently went to 0 while the others passed.

## CI at STEP level, on the rebased head

Run **`34508981106`**, workflow CI, event `pull_request`, head_sha `13a6de71…` — asserted equal to
the sha I pushed. Read at **step** level, because a job conclusion aggregates and a step conclusion
does not.

| job | steps | not success/skipped |
|---|---|---|
| `clippy` | 8 | 0 |
| `release-invocations` | 7 | 0 |
| `test` | 9 | 0 |
| `python-ast-tests` | 9 | 0 |
| **total** | **33 across 4 jobs** | **0** |

`cargo clippy --all-targets -- -D warnings` = success, genuinely run. `cargo test` = success.
`scripts/ast tests must actually run` and `scripts/ast/test_*.py` = success. The run conclusion is
`success` too, but the 33 step conclusions above are the evidence.

### Which tree CI actually tested

A `pull_request` run builds the **merge ref**, not the head, so the deterministic check is the tree
rather than the async `mergeable` field:

| | |
|---|---|
| `refs/pull/156/merge` | `7b74724b3e7624aabaf0a3bf22bd60e2082a1ff6` |
| its parents | `4bc5c1cf…` (main) + `13a6de71…` (this head) |
| **merge tree** | **`05e165c8abdc7405ff46fb0a7bb1116ce9bc562c`** |
| head tree | `05e165c8abdc7405ff46fb0a7bb1116ce9bc562c` — **equal** |
| main tree | `7606210893816ec83c277e79adc822d417e0114b` — different, so the run tested the change |

`behind_by` is 0, so merging into main is a fast-forward and the merge tree must equal the head
tree. **That merge tree is also exactly what `git merge-tree --write-tree` predicted before the
rebase was performed** — a pre-registered prediction, recorded in the lane doc at G0.4 and hit.

`mergeable=True`, `mergeable_state=clean`, read 2026-09-10T17:36:59Z — an asynchronously computed
field, recorded with its timestamp and **not** relied on. The tree equality above is the claim.

## Registered limitations

1. The fix makes the config-helper branch join the existing cross-file pass. **Whether that pass
   accepts a file-shaped target for other helpers is out of scope here.**
2. **`N-EXTRACTOR-SHADOW` still stands.** A user running `base sync --ast` gets the **global**
   extractor, not this one. This PR does not change that; the release will, by shipping the new
   extractor in the asset. Every reading above names the script path it ran, and none is the global
   copy.
3. The reconciliation with the earlier build's 14 is an inference about a corpus that no longer
   exists, marked as such above.

ChristopherKahler:
**REVIEWER G2 VERDICT — `petrel`: PASS.** Two findings, neither a blocker. `auk` merges; I do not.

Graded head `13a6de718fbd6e6f203148cb93ba4c5a26e08f15`, frozen throughout (API reads 17:53:53Z and 18:08:41Z, unchanged). Every number here is mine, from my own worktrees, my own corpus export and my own instruments. My conditions D1-D16 were written to `preground-156.md` **before** I opened the build record, the result sections, this comment or any CI page.

### The measurement, my own arms

| reading | before (`main 4bc5c1c`) | after (`head 13a6de71`) |
|---|---|---|
| `uses_config` EDGES (detector B) | **0** | **13** |
| `usesConfig` token occurrences (detector A) | 0 | **26** = 13 plain triples + 13 `<< >>` annotations, **0 fitting neither shape** |
| distinct `ops:` predicates in the TTL | 20 | 21 |
| ambiguity reports on stderr (leg 9) | **0** (the leg's own control) | **1** |
| files the extractor collected | 242 | 242 |
| distinct edge targets / sources | - | **1** (`config/model.php`) / **13** |

Leg 9's line, verbatim from my own head arm:

```
base-ast: uses_config target 'app.php' at .../app/Http/Middleware/HandleInertiaRequests.php:42 matches 2 nodes (bootstrap_app_php, config_app_php) - ambiguous, no uses_config edge emitted
```

`app.php` appears in **0** emitted edges: reported, not silently resolved.

### The gates that decide it

- **The G0.4 prediction was RECOMPUTED, not dated.** `git merge-tree --write-tree main 638fe2e5` reproduces `05e165c8abdc7405ff46fb0a7bb1116ce9bc562c` with **0 conflicted paths**, so whether that value was written before or after the rebase stops mattering: it is a repeatable measurement, not a claim.
- **Leg 7, plus the leg blob identity cannot give.** 3 of 3 blobs identical; file counts 292 -> 292 and `scripts/ast` 19 -> 19; `name-status` shows only the 11 base-catch-up files; the `scripts/ast` zero sits behind a **309**-line positive control and a **700**-line same-pathspec canary.
- **The tree, never a sha and never `mergeable`.** `tree(head)` = the merge ref's tree = `05e165c8...`, and differs from `tree(main)` `76062108...`. `condor`'s `mergeref.py` positive control read main's tree and its **404 canary fired**.
- **CI at step level**, run `34508981106`: 4 jobs, **33 steps visited**, zero not-success, zero skipped, every job on the graded head. `ci.yml` is byte-identical at head and main; `continue-on-error` appears once, at line 128, **inside a comment**; `cargo clippy --all-targets -- -D warnings` = success.
- **Leg 9 survives a warm cache**, reproduced independently: 157 of 242 uncached on the second run, report still 1, TTL byte-identical cold vs warm.
- **The must-fail canary for the before column.** `main`'s own extractor emits **2** `uses_config` edges on an in-file fixture, so its **0** on the real corpus is a real absence and not blindness.
- **Tests:** main 9 of 9, head 10 of 10, all rc 0. `ran 10 of 10 legs` against 10 `^def test_`. CI's own job log carries 6 `ran N of M` lines with no N != M.
- **Custody:** the client tree is byte-identical before and after my run (HEAD, dirty count 0, 117 cache files, cache-list md5 unchanged).

### The leg neither the builder nor the orchestrator ran

Both measured the 242-file client corpus, which cannot separate *the fix works* from *this corpus happens to contain `config/model.php`*. On a **3-file fixture**, in two separate worktrees:

| shape | main | head |
|---|---|---|
| same-file (the in-file path from #118) | **2** | **2** |
| cross-file (the Laravel shape) | **0** | **2** |

Cross-file **0 -> 2** with no client corpus involved, and the in-file path does not regress.

### Finding 1 - not a blocker

The commit added after the previous reviewer's pre-ground (`5d9b0a32` -> `638fe2e5`) changed **`scripts/ast/relation_vocabulary.py`** (+136 -28), blob `1070ac6070c17044ae35b8c6a4a52397b6ceaea0` -> `ffda1c9a799398c9d1582ce45bf391176321c8ed`. The other two files are byte-identical across it. Continuity was re-established for `extractor.py` against the original builder's sha256, and not for the one file that actually moved. It does not qualify the PASS - every gating number was re-taken at the rebased head by three independent parties - but no future reader should carry the original builder's green forward for that file.

### Finding 2 - for the backlog, not this PR

`resources/views/app.blade.php:7` calls `config('app.name', 'Laravel')`. It is outside every population measured in this lane, twice over: the census scope was `app bootstrap routes config database tests`, and the file is `.blade.php`, which has its own extractor arm. Not a defect here and not in scope.

### One artifact of my own, disclosed

My first population pass read **25** references against the record's **26**. Before reporting a discrepancy I asked what each instrument counts: my pattern required a closing paren straight after the quote and so could not see `config('model.pricing', [])` at `app/Services/Ai/ModelCallLog.php:98`, a call with a second argument. **The record's 26 is correct; my 25 was my own artifact.** With it restored the population closes exactly: **26 = 13 edges + 1 reported ambiguous + 12 collapsed**.

### After the merge

`main` is now `5ef2be25e365eac63ee423fe789e1e7c57afdc96`, a squash whose **tree is `05e165c8abdc7405ff46fb0a7bb1116ce9bc562c` - exactly the tree that was graded** - with a single parent `4bc5c1cfbc1cef919cc4474f9148f6e6b4aa5fd0`, all three files at their graded blob shas, and **0** files changed outside `scripts/ast/`. The graded head is not an ancestor of `main`, which is correct for a squash merge and is not a failure.

PR #167, opened 2026-09-19
base 0.16.0 lane 3: deferred state for records and projects, rank 09 migration removed, operator pointer

ChristopherKahler:
Lane 3 of base 0.16.0: deferred state for records and projects, the rank 09 migration **removed**, and the operator pointer that makes deferral discoverable.

Merged with `main` (`f433a9a`) at `0488837`. 19 commits.

---

## What this branch does

**Records and projects that go cold stop surfacing at session start, and can be brought back one at a time.** Handoffs, forks, tasks and milestones defer on their own clock; projects defer on their folder's newest file modification time. Nothing is deleted — every deferred thing stays listed, queryable, and one command from coming back.

## The largest change is a REMOVAL

Rank 09 shipped a migration gate: while an "upgrade migration" was pending, the deferral pass withheld every write, so a legacy install could not mass-defer before its operator had previewed. That gate, its machine-wide marker, and `mark_fresh_install` are **all gone**.

The reason is design intent that the original design lacked: **deferring is not deleting.** A deferred record stops surfacing, stays listable and queryable, and every listed row prints its own revive command. So the day-one sweep the gate existed to prevent is the feature working correctly — the gate was blocking a correct outcome, and the scope defect that blocked this branch's merge was a defect in a mechanism that should not exist.

### Three consumers, each resolved explicitly

| site | what happened |
|---|---|
| `protocol/reconcile.rs` | the gate itself. `migration_pending` gone; `writable` is now unconditional on both `Defer` and `Revive`. |
| `hook/session_start.rs` | the `"NOTHING HAS BEEN WRITTEN. N records are waiting on the upgrade migration"` stderr line is gone. It announced a block that no longer exists. |
| `doctor.rs` | **both** migration warnings removed. |

### Three losses stated rather than absorbed

1. **The marker is now write-only in production.** `base defer migrate --apply` still records it and `--rollback` still clears it; nothing outside the tests reads it.
2. **No operator-visible surface states that a migration ran.** `doctor`'s `Applied` branch was the only one and **nothing replaces it.** Its text also advertised `base defer migrate --rollback` as the undo, which is false at any distance from the run — `doctor::restore_tier` replaces whole tier files, so an undo on day 11 silently discards every unrelated write since.
3. `doctor`'s `Pending` branch could not survive either: with nothing writing the marker at install, an absent marker no longer separates "upgraded, not migrated" from "installed yesterday, nothing to migrate", so it would have told every new 0.16.0 user, forever, that a migration was pending.

**`base defer migrate` survives** as an opt-in way to reset `lastActive`. Resetting legacy clocks is a real operator action that never depended on the gate.

## The pointer

Session start already printed a count. It now also tells the operator what that count means and how to act on it — **two lines, one per population, with deliberately different wording.**

```
base: defer — 4 deferred, 0 revived (4 records scanned)
  Deferred records are not deleted. They stop showing up, and they stay listed.
  That count covers four kinds, and each one lists separately:
    base handoff deferred     base task deferred
    base fork deferred        base milestone deferred
  Every row prints the command that brings that record back.
  Opening the file does NOT bring one back. Only that command does.
base: reconcile — 1 deferred, 0 revived (1 projects scanned)
  Deferred projects are not deleted. List them with: base project deferred
  Every row prints the command that brings that project back.
  A PROJECT defers on its FOLDER, not on its record. base reads the newest
  file modification time under the project's path, skipping .base, .git,
  node_modules, target and the like. Working on the project inside base does
  not count here. Changing a file under the folder does.
```

**Why the last line of the record block cannot be left out:** reading a deferred handoff moves its clock but does **not** revive it. Opening the file looks like it should work and silently does not, and this is the only place an operator is ever told which command actually works.

**Why the project wording differs:** a record defers because nothing *edited* it; a project defers because no *file under its folder* changed. `protocol::touch::IGNORE_DIRS` contains `.base` by name, because `graph.nq` is rewritten every session — so an operator who edited the project record inside base has written into a directory the walk refuses to enter. Identical wording would be true for one population and false for the other.

These are stderr lines, which are outside every output budget by construction (`src/emit/mod.rs`): Claude Code feeds only a hook's stdout to the model. Nothing here competes with a signal block.

## Verification

**Full suite on the merged tree: 90 binaries, 1157 passed, 7 failed, 4 ignored.** All seven failures are pre-existing and belong to other lanes — five in `session_start_layout_test.rs` and two `graph::tests::purge_*` date-sensitive tests — each attributed by full name, with a fix already existing on the lane-1 branch.

**The removal was verified by mutation, not by the suite going green.** The bar: run one mutation (invert the staleness comparison in `plan_records`), record which tests redden **by name**, rewrite the fixture, run it again. Any name that reddened before and not after is a test the rewrite weakened.

| | names | AX1 red |
|---|---|---|
| before | 40 | 20 |
| after | 35 | 18 |

Five names left the reddened set: three renamed and still red under the same mutation, two deleted and **absent** from the after-run rather than merely green. Three names joined it, all three the new names of those renames. **Tests weakened: zero.** 20 − 5 + 3 = 18.

**The dangerous part of this change was a test helper, and it was caught by mutation.** `migrate_test`'s `pending()` read the string `"deferral upgrade migration PENDING"` out of `base doctor` output — which this branch deletes. Left alone it becomes a constant `false`: four assertions would have gone red and been noticed, but two `assert!(!pending(..))` sites would have **passed no matter what the migrate command did**. It now reads the marker directly, and a mutation making `state()` misreport proves it still discriminates: exactly those two red, nothing else, all four positive sites green.

**Five tests were deleted, by name, each forced by the removal** — two asserting a cold record is *not* deferred while the marker is absent (their required answer inverted, it did not weaken), one asserting the gate itself, and two calling `mark_fresh_install` directly. One test was required to survive and did, renamed: `a_cold_record_defers_and_a_revivable_one_revives_in_the_same_tier`, which catches a `Defer` filter stuck *on* — the most likely way this removal goes wrong. A mutation dropping `Defer` at the write filter reddens it along with 12 others.

**`mark_fresh_install` is absent from the tree as code:** 8 call sites before, 0 after, with a positive control on the same pattern (3 live `mark_applied` hits) and a second control proving the same search found all 8 at the old head. Nine mentions survive in comments **on purpose** — every removal site keeps a headstone naming what went and why, so a reader who remembers the function discovers it was removed rather than finding a file that never had it.

**Every command the pointer names was executed** against a purpose-built base home through the real hook, and all five returned records.

## Merge notes

Merged rather than rebased, deliberately. Every commit on this branch was graded individually — index equal to worktree per file by blob hash, then three independent tree readings required to agree — and those commit SHAs are cited with their tree hashes across the round's documents. A rebase would have rewritten all 19 and voided that record.

Two files conflicted. `src/crud/handoff.rs`, where both lanes landed the same tier-walk refactor in different homes; lane 2's richer `create` wins and the shared helpers now live in `crud/mod.rs`. And the base-help CLI reference, whose conflicted block is a transcript of `base handoff --help` — resolved by reading the block out of the merged binary rather than splicing the two transcripts by hand.

A test parked with a prediction settles here: `a_new_handoff_archives_the_older_deferred_one` was red on this branch and is green after the merge, which is what the parking note predicted. It was shown red first, so the green is a measurement rather than an assumption.

## Not in this PR

No preview at day 11, no bulk undo, no lazy fill, no window change. Each was measured and dropped.


ChristopherKahler:
## Updated: lane 1 is now integrated into this branch

The branch has moved from `0488837` to `3a1915f`. The PR description above was written before the lane 1 integration and describes lane 3 alone.

**The whole suite is green: 92 binaries, 1182 passed, 0 failed, 4 ignored.** The seven failures the description calls "pre-existing and belonging to other lanes" are fixed, not excused.

### What was added, and why it is four commits rather than a merge

`plover/0160-budget-and-layout` holds nine commits. Comparing each against this branch's, patch text to patch text, **five are the same change at different line offsets** — four byte-identical, one differing by a single documentation line. A merge presented 86 conflict hunks across 15 files, and almost all of them were this branch's own work disagreeing with itself about line numbers.

Only four commits are genuinely lane 1's, and all four applied with **zero conflicts** — which is itself the evidence that the 86 hunks were not two competing implementations:

| | |
|---|---|
| `649ed30` | rank 04 — the memory block gets its own budget, in whole notes |
| `75721fb` | DUE NOW never collapses, not even for the first screen |
| `e19109c` | the purge fixture is seeded relative to now, never absolute |
| `a33f827` | rank 10 — doctor reports what each hook emitted |

Authorship stays with plover on all four. What is lost is the branch topology — a record of *how* this arrived, not *what* arrived. The missing merge parent should not be read as lane 1 having been skipped.

Each was confirmed present by content afterwards, not by trusting cherry-pick's exit code: the files those commits create exist, the DUE NOW inversion is present and its predecessor absent, and the purge fixture no longer carries an absolute date.

### The one interaction the integration produced

`3a1915f` changes one literal in `tests/session_start_layout_test.rs`. Lane 1's first-screen test expected session start to print `clear: base reminder remove <slug>`; this branch prints `archive`.

That is not a rename. Rank 06 added `base reminder archive` beside the existing `remove` and changed which one session start **recommends**. Both still exist:

```
archive  Archive a reminder: it stops surfacing and is kept. `remove` deletes
remove   Remove a reminder (hard delete)
```

The old line told an operator to hard-delete a reminder they had merely handled. That is the suggestion rank 06 replaced deliberately.

So the test's expectation was stale and the test moved. The assertion is untouched — it still pins DUE NOW's last item inside the first 2,000 UTF-16 units, which is what the test is actually for. Changing the output instead would have bent production to a stale literal and restored a destructive recommendation.

`Cargo.toml` remains at `0.15.2`. No release has been cut.

PR #168, opened 2026-09-21
base 0.16.0 lane 1: ranks 00, 03, 04, 10 — session start emission, the lastActive clock, the memory budget, and doctor's sizes

ChristopherKahler:
# base 0.16.0 lane 1 — ranks 00, 03, 04 and 10

Nine commits on `f433a9a`, zero behind. This is the lane that makes every other
rank in 0.16.0 **observable**: today nothing base emits at session start reaches
Claude past the first 2,000 characters, and the notice of that loss is itself
lost inside it.

**`Cargo.toml` stays at `0.15.2`. This PR does not cut the release.**

## The four ranks

| rank | the defect, as the work order states it | commits |
|---|---|---|
| **00** | Nothing base emits at session start reaches Claude past the first 2,000 characters, and the only notice of the loss is itself lost. `signal/mod.rs:156-158` pushed `[+N signals suppressed — budget cap]` onto the tail of `combined`, at byte ~42,000 of 46,281 — inside the part the host truncates. | `9679330`, `a3231ff`, `7f4a83e`, `fc4c2f4` |
| **03** | The `lastActive` clock has never ticked on any handoff, in either tier. Every age base reports is really an age since creation, and Part C's whole deferral feature is built on that field. | `2e0d6a0`, `167c714` |
| **04** | The memory signal loads ~970,366 characters of notes on every session start and discards 99.6% of them after the cost is paid. It is a second guard that looks like a budget and is not. **Fixing rank 00 without this makes session start worse, not better**, because the cap that hides it is the cap rank 00 removes. | `176d5ac` |
| **10** | `base doctor` cannot report what base emitted or whether it was trimmed. Without it nobody can tell whether rank 00 stayed fixed. | `9679330`, `e6a099b` |

## Every commit in this PR

| commit | what it does | rank |
|---|---|---|
| `2e0d6a0` | `refactor(crud)`: share the tier file walk with the post-tool hook. | 03 (prerequisite) |
| `167c714` | `fix(hook)`: the `lastActive` clock ticks for handoff and fork docs. | **03** |
| `9679330` | `feat(emit)`: measure a hook's output before it is printed. | 00 / 10 |
| `a3231ff` | `feat(emit)`: route session start through one measured emission. | 00 |
| `7f4a83e` | `feat(session-start)`: header, instructions, then blocks by rank. | 00 |
| `176d5ac` | `feat(memory)`: give the memory block its own budget, in whole notes. | **04** |
| `fc4c2f4` | `fix(emit)`: DUE NOW never collapses, not even for the first screen. | 00 |
| `ffd223b` | `fix(test)`: seed the purge fixture relative to now, never absolute. | **none — a pre-existing defect on `main`, see below** |
| `e6a099b` | `feat(doctor)`: report what each hook emitted, and the legacy `max_chars`. | **10** |

34 files, +6,336 / −505.

## Rank 03's figure, stated exactly

**339 records carry `handoffDoc` across both tiers, and all 339 had `lastActive`
exactly equal to `createdAt`. Zero had moved.** Workspace tier 275 (176 `"fork"`,
99 `"handoff"`), global tier 64 (43 `"fork"`, 21 `"handoff"`).

**There is no `Fork` type in this graph.** `rdf:type ops:Fork` returns population
zero in both tiers. Forks and handoffs are one record type separated by a `kind`
literal, which is why **one predicate fix covers both** and why this is one code
path rather than two.

An earlier draft of `167c714`'s message put the `path`-predicate control at "74
records moved". That figure did not reproduce — a re-measure returned 32 against
a different filter — and it is **withdrawn rather than restated**. The control
establishes a direction, not a count, and the direction is all this PR claims
from it.

## `ffd223b` is a PRE-EXISTING DEFECT ON `main`, carried by this PR — not lane 1 work

Called out because a reviewer reading the diff would otherwise attribute it to
this lane, and because it explains why every lane's suite was red.

`purge_fixture` in `src/graph.rs` seeded its "fresh" note with an **absolute
date, `2026-06-18`**, while `purge_stale` builds its cutoff from `Local::now()`.
The fixture therefore aged. **It aged out on 2026-09-16, and the window in which
it passed was one day wide.** From that date it failed every lane's full suite
and every pull request's CI, on `main`, for reasons unrelated to whatever that
pull request changed.

It was assigned to this lane only because lane 1's PR is the next thing reaching
`main`. The fix seeds relative to now.

**Proven by the flip, not by a green suite.** On `src/graph.rs` alone, both purge
tests fail at the parent commit and pass with the fix, while the two other purge
tests the same filter catches stay green in both arms — an unplanned independence
control. A sweep of the tree found **108 absolute date literals across 191 `.rs`
files**, and exactly one on the fresh side of a live cutoff. The first sweep
reported zero and was wrong, because an ERE `\?` matches a literal question mark;
a control caught it and the sweep was re-run.

## One registered test name is SUPERSEDED, not fixed — read this before grading the merge

`fc4c2f4 fix(emit): DUE NOW never collapses, not even for the first screen` is a
**deliberate behaviour change**, and it states that on the tin.

`tests/session_start_layout_test.rs` is this lane's file (+508 lines here). The
older copy on another branch carries
`due_now_collapses_only_to_keep_the_first_screen_and_the_header_counts_it`.
**This PR does not fix that test. It deletes it and replaces it with
`due_now_stays_whole_and_first_when_it_overflows_the_first_screen`, which asserts
the inverse.**

**So that name will never go green — it will cease to exist.** Any grader keyed
on it must record it as *superseded by the `fix(emit): DUE NOW never collapses`
commit, replaced with the inverse assertion* — never as fixed, and never given a
SHA that reads as a measured cause.

Measured on this branch, both controls: the old name is **absent** (negative
control: a name known not to exist also returns 0; positive control: the other
four registered names each return exactly 1 `ok` line).

## Open item this PR does not close: J8

`docs/BASE-WORK-ORDER.md:442-450`.

Rank 10 records the emitted size of **session start only**, because that is the
only hook event that currently goes through the measured emission. `src/config.rs`
says `prompt_chars`, `pre_tool_chars` and `post_tool_chars` "are read by nothing
yet".

So "record the emitted count per hook event" is **parked, not narrowed**, for
user-prompt-submit, pre-tool-use, post-tool-use and stop. Each gets its record,
and its row in the design-to-code map, in the J8 commit that routes it through
`Emission`. Stop's only output today is a `systemMessage` for the operator, and
the spec budgets Stop only "if it emits model context", so **J8 also rules whether
Stop gets a record at all.**

## Risk a reviewer should know about: the Windows rename, and which way it fails

**The mechanism.** `emit::record::keep_capped` (`src/emit/record.rs`) renames
`hook-output.jsonl` to `hook-output.1.jsonl` once it reaches 1 MiB, then appends.
The rename goes through `store::rename_with_retry`, which retries **only** the
transient lock family, 5 attempts with 15+30+60+120 ms of backoff, so a conflict
has about 225 ms to clear. The bound is deliberate: every caller sits on a path a
hook reaches on every tool call, so an unbounded spin would turn a stuck file
handle into a hung command.

**Windows is where this gets exercised.** Replacing a file that another process
holds open is refused on Windows in cases where POSIX replaces it atomically, and
this file is one Claude reads.

**THE DIRECTION, which is the part worth acting on: it fails toward an oversized
file, never toward a lost record.** If the rename is still refused after the retry
budget, the error is deliberately discarded, **the append still happens**, the file
grows past its 1 MiB cap, and the next record retries the rename. The reader reads
both `hook-output.1.jsonl` and `hook-output.jsonl`, so `base doctor`'s 20-run
window stays correct throughout.

**So a reviewer investigating this should look for a `hook-output.jsonl` larger
than 1 MiB — not for missing runs, and not for an error in the logs.** There will
not be one. Nothing is lost and nothing is reported.

**CI cannot tell you whether this happens.** Every `runs-on` in
`.github/workflows/ci.yml` is `ubuntu-latest`, and `cargo test --no-fail-fast`
runs in one of those jobs. `release.yml` builds on other platforms, but its only
`cargo test` is also in an `ubuntu-latest` job. **No workflow runs the suite on
Windows, so a green CI on this PR is not evidence about the rename path.**
`rename_with_retry` carries its own unit tests, including one that it does not
retry a non-transient error, and those do run — on Linux.

## Rank 10 can report VOID, which is the point of it

`base doctor` prints **"no run on record"** where session start never ran —
**never a run of zero**. An instrument that cannot say *"I could not see"* turns
a failed measurement into a passing one, and rank 00 is exactly the defect where
that matters: the loss report died inside the loss.

The record is written **before** the print, so a failed session start is still on
record. Append-only, rotated at 1 MiB, deliberately kept out of
`hook-events.jsonl` because that file keeps only its last 5,000 lines and session
starts were **0 of the last 5,000** in a live global log.

## How this was verified

Green window on the tip, one run, all four legs, tree locked before the first leg
and re-read after the last:

| leg | result |
|---|---|
| toolchain asserted present first | cargo 1.95.0, rustc 1.95.0, clippy 0.1.95 — so a 127 could not read as a pass |
| `cargo clippy --all-targets -- -D warnings` | rc 0, **0 diagnostic lines** |
| `cargo test --all-targets --no-fail-fast` | rc 0 — **89 suites, 1,140 passed, 0 failed, 3 ignored** |
| tree lock | `bd018a69…` before **and** after, porcelain 0 |

The permissive clippy arm runs so a could-not-compile cannot be misread as a lint
failure. Failures are graded **by name**, never by count: a count cannot tell a
registered failure from a different one.

Earlier in the lane: eight mutants, each on the green tree, each returning the
tree lock before and after its mutation, every one matching its registered
prediction. Two unregistered failures, both classified by a stated rule rather
than by opinion — does the panic name the field the mutation corrupts? **One
mutant found a hole in rank 10's own test:** a test asserting that doctor tells
the last run from the largest stayed green while the `largest` field was broken,
because its fixture made run 1 the larger one by construction. The fixture was
fixed rather than the prediction re-registered, and a control was added so it
cannot drift back.

## Two commits were reworded before this PR opened, message-only

`petrel` found two wrong claims in commit messages. Both are corrected in place
rather than left standing, and **every tree on the branch is byte-identical
before and after**:

- `2e0d6a0` said `apply_to_tiers` was "untouched". Two hunks of that diff sit
  inside it and inside `searched_tiers`. The message now lists, per function,
  what changed: two functions untouched, three changed by the same call-site
  rename, five lines in total.
- `167c714` carried the unreproduced "74 path records moved" figure, now
  withdrawn as described above.

Old-to-new map, all nine, each pair sharing one tree object:

| old | new | tree |
|---|---|---|
| `6e9495f` | `2e0d6a0` | `293aa517…` |
| `fee00ae` | `167c714` | `83ddb641…` |
| `180e57d` | `9679330` | `5f86fdaf…` |
| `6a5e5f6` | `a3231ff` | `d104eab4…` |
| `357c65b` | `7f4a83e` | `8d2abd00…` |
| `15744bd` | `176d5ac` | `9fd9838e…` |
| `8122724` | `fc4c2f4` | `ed7a2a17…` |
| `21b4ae8` | `ffd223b` | `634ef0ec…` |
| `8a75593` | `e6a099b` | `bd018a69…` |

The whole-range diff `f433a9a..HEAD` is **byte-identical** before and after the
reword (338,989 bytes), with a negative control proving the comparison can see a
difference when one exists. The green window above ran on tree `bd018a69…`, which
is still the tip's tree, so it measures this PR and not a superseded one.


ChristopherKahler:
Closing as **superseded**, not abandoned — and measured rather than assumed.

Seven of this stack's thirteen commits were already on `main` by `git patch-id --stable` before any rebase began, measured twice independently with the same seven each time. Of the rest, four landed on `main` as `13ea6cc`, `848e434`, `621dfef` and `77c6839`, and one produced **zero changes** on cherry-pick because its content had arrived by another route: `main` already had `all_tier_files` and `tier_label_of_file` in `crud::mod`, and gone from `crud::handoff`.

**That leaves `7f4a83e`, the session-start layout half, and `main`'s `session_start.rs` supersedes it entirely:**

| | this PR | `main` |
|---|---|---|
| functions absent from the other | 0 absent from `main` | has all 17, plus `render_adhoc_queries` |
| LAYOUT kinds | 30 | all 30, plus `working-set-scope` (31) |
| `instruction_block` | `(letters)` | `(letters, deferred)` |
| `header_line` | `(facts)` | `(facts, deferred)` |
| budget notice unit | units, `session_start_chars` | bytes, `session_start_bytes` |

Both of `main`'s signatures are strict supersets, and the budget notice here is in the retired unit.

Confirmed live on the merged `main`, not merely present in source: session start renders the header line and the honest-absence line, `WORKING SET: no graph was read ... This is NOT an empty working set.`

Nothing in this PR is lost by closing it. `main` is `2d3803babcc5db38e5517d205a03c62ecf31e835`.

PR #169, opened 2026-09-21
base 0.16.0 cross-lane item 2: the working set reads both tiers

ChristopherKahler:
# base 0.16.0 — cross-lane item 2: the working set reads both tiers

**Stacked on #168.** Base branch is `plover/0160-budget-and-layout`, so the diff here is
one commit. GitHub retargets this to `main` when #168 merges.

## The defect

A task or milestone recorded in the **global tier** never rendered in the active-awareness
block. `run_sections` built its query and handed it to `crud::load_and_query`, which opens
the workspace graph alone, so a global-tier record could not match.

**The block did not say a tier went unread. It showed a shorter list.** That is the shape
where absent reads as empty, and it is the one nobody reports, because a short list looks
like a correct answer.

## The fix, and why it is a sibling rather than a change

`crud::load_merged_and_query` is added **beside** `load_and_query`, not in place of it.
Changing the shared function would alter behaviour at every remaining call site at once.
`active_awareness` is the new function's first and only caller.

The seam already existed. `store::load_merged` merges both tiers, returns `None` only when
neither graph exists, and already degrades to a lenient per-tier load when one tier will not
parse. `flow_resurface` — same directory, same kind of block — has called it at three sites
all along. `active_awareness` is the one signal that never moved onto it.

`store::query`, not `store::query_union`: base's own queries wrap their patterns in
`GRAPH ?g { … }`, so they match named graphs explicitly. `query_union` would return the very
same rows, because an explicit `GRAPH ?g` is unaffected by a union default graph — but it
would tell every later reader that this query was user-authored, which is false.

## What this deliberately does not do

The **absent envelope** is a separate item. When neither tier has a graph this errors and
says so, rather than returning zero rows. It does not yet render a block naming which tiers
were read and which degraded. Absent, empty and zero are three states; only the first two
are separated here.

## The remaining single-tier call sites, with the unit stated

The previous count of these does not reproduce, so here is the measurement rather than the
number alone. **Unit: lines under `src/` containing `load_and_query(`, excluding the
definition and excluding doc comments, a multi-line call counted once at its opening line.**

| directory | sites |
|---|---|
| `crud/` | 29 |
| `signal/` | 5 |
| `domain/` | 6 |
| `install.rs` | 1 |
| **total** | **41** (42 before this commit moved one) |

The design doc recorded "30 call sites across `crud/` and `signal/`". Those two directories
measure **34** here and **35** before this commit. That figure is not reproduced, so it is
not restated and not silently replaced. Every one of the 41 reads a single tier and carries
the same limit latent. **This commit fixes exactly one of them.**

## Tests, predicted before they ran

| arm | prediction | result |
|---|---|---|
| `a_global_tier_task_renders_in_the_working_set` | red before, green after | matched |
| `a_workspace_task_still_renders_when_both_tiers_exist` | green both sides | **missed, see below** |
| `a_workspace_only_task_is_unaffected` | green both sides, control | matched |
| `an_entity_recorded_in_both_tiers_renders_once` | green both sides, control | matched |

**The miss is recorded, not edited.** The second arm was registered green on both sides as a
guard that merging does not lose the workspace tier. I then wrote it with a second assertion
— that the global row is present too — which makes it red before the fix by construction.
Its stated purpose did hold: the pre-fix failure printed the workspace row and named the
global assertion as the one that failed. The extra assertion is kept; the prediction is left
wrong on the record.

The fourth arm pins behaviour that is **currently right by accident.** `run_sections` keys a
`HashMap` on entity id in order to gather an entity's several owner links into one row.
Merging the tiers is the first thing that makes two rows share an id, so that `HashMap` is
now what stops an entity recorded in both tiers rendering twice — for a reason unrelated to
why it was written. Nothing else protects it and nothing announces if it breaks.

Each arm takes its own fake home through `home::with_thread_home`, because `load_merged`
resolves the global tier through `home_root()` and this test binary's process-wide root is
shared by every thread in it.

## Verification

Full green window on the committed tree, tree locked before the first leg and re-read after
the last, and the commit asserted to **be** that tree:

| leg | result |
|---|---|
| `cargo test --all-targets --no-fail-fast` | rc 0 — **89 suites, 1,144 passed, 0 failed, 3 ignored** |
| tree lock | `d4028871…` before, after, and as `HEAD^{tree}` |

1,144 is 1,140 on #168's tip plus exactly these four arms.


ChristopherKahler:
Landed on `main` as `13ea6cc` — cherry-picked rather than merged as a branch, which is why GitHub did not close this automatically.

`main` is now `2d3803babcc5db38e5517d205a03c62ecf31e835`. Verified before the push: 96 targets, **1,215 passed, 0 failed, 5 ignored**; clippy `--all-targets -D warnings` rc 0 with the permissive control also rc 0. `relay_wake_script_test` is excluded and named, not green — it hangs on Git Bash (192s and 292s on earlier measurements, past 25 minutes this round).

Closing: the content is on `main`.

PR #170, opened 2026-09-21
base 0.16.0: the honesty envelope — the working set says what it read, and what it did not

ChristopherKahler:
# base 0.16.0 — the honesty envelope: the working set says what it read, and what it did not

**Stacked on #169.** Base branch is `plover/0160-crosslane-item2`, so the diff here is two
commits. GitHub retargets this to `main` as the stack merges.

| commit | what it does |
|---|---|
| `96b6744` | `fix(store)`: hand back what a merged read managed to read, as a value. |
| `bea962c` | `feat(signal)`: the working set says what it read, and what it did not. |

Two commits and one PR, because the first exists only to make the second possible. Shipping
it alone would publish a commit whose reason is not in its own pull request.

## The defect

Absent, empty and zero are three different states. Two of them put the same thing on the
operator's screen: **nothing**.

These are the two arms in `src/signal/mod.rs` that decide it:

```rust
Ok(_) => diagnostics.push(format!("<{hook}-active-awareness:no-match>")),
Err(e) => eprintln!("base: signal 'active-awareness' failed: {e}"),
```

Neither pushes a block. A graph that could not be found took the `Err` arm and left a line on
stderr. A graph that was read and held nothing took the `Ok` arm and left a diagnostics tag.
**No working-set block reaches the operator either way**, so *"I could not read your graph"*
and *"you have no work"* render as one screen. They are opposite claims.

## Why the first commit had to come first

`load_merged(cwd) -> Option<Store>` reported a damaged tier by printing to stderr and returned
the same `Some(store)` either way. **The signature settles it**: no caller could learn that a
tier was skipped, whatever the body did. Evidence that leaves as a side effect is invisible to
the code that has to decide what to tell the operator.

`load_merged_reporting` returns a `TierRead` beside the store. `load_merged` keeps its name,
its signature and its stderr line as a thin wrapper, so **all 13 of its call sites are
untouched**.

### A second defect, found while reading the first, and worse than it

The stderr warning was gated on the malformed-**line** count:

```rust
if total_bad > 0 { eprintln!("graph: skipped {total_bad} ...") }
```

A tier that cannot be **opened** takes a different arm, contributes no lines, and never
touches that count. So when every tier fails to open, the count is zero, the gate stays shut,
and `Some(store)` comes back with nothing read into it and nothing said.

**Measured on the unfixed code**, with a throwaway probe that was removed afterwards, tree
clean on both sides:

```
load_merged returned Some(store) holding 0 quads, from two tiers NEITHER of which
could be opened. Nothing in the return value distinguishes this from a graph that
is genuinely empty.
```

`TierFault::skipped_lines` is now `Some(n)` when the tier loaded and n lines were skipped, and
`None` when the tier could not be opened and contributed nothing. Summing those back into one
number is how the old gate missed its own worst case, so `is_clean()` is the health question
and `skipped_lines()` says in its own doc comment that it is not.

**One deliberate behaviour change to a shared function**: the warning now fires on a fault
rather than on a line count, so `load_merged` prints in one case where it printed nothing
before. That case is the bug above. It is stated rather than buried, because "no output" was
never a contract anyone wanted.

## What the block now says

| state | what renders |
|---|---|
| no graph in either tier | `WORKING SET: no graph was read … This is NOT an empty working set.` |
| a tier read badly | the scope line, plus `AND THIS READ WAS INCOMPLETE: <what went wrong>` |
| everything else | `SCOPE: workspace 'x' and the global tier; other workspaces were not read` |

The scope clause is on **every** render, not only on not-found. It is what makes an empty
working set safe to show at all: without it, "nothing in scope" and "nothing exists" are the
same sentence, and only the first is one the code can honestly make.

## A constraint read off the tests before designing, not discovered after

`tests/session_start_layout_test.rs` matches the `PROJECTS`, `TASKS` and `MILESTONES` headers
with `l == whole`, an exact line comparison. Widening a header to carry the scope clause would
have satisfied the honesty contract by breaking the layout one.

So the clause is **a line of its own**, and every existing header is byte-identical. The
comment at the push site records why, so nobody later tidies it into the header.

## A deliberate contradiction of #169's description, named rather than left

`crud::load_merged_and_query` returned `Err` when neither tier had a graph, and **#169's body
states that as a feature**. It now returns `Ok(None)`.

The `Err` was the right placeholder while nothing could render the difference — refusing to
answer beats answering "zero rows" wrongly. Now that a caller can render it, a state the
caller can act on beats an error string it cannot. #169 is accurate about the commit it
describes; this commit changes it, and saying so is cheaper than letting two published
descriptions drift apart.

## What this does not do

**State 3, the workspace axis, is named and not closed.** Other workspaces are still never
opened, on every call. The scope clause says so out loud rather than leaving a reader to
assume otherwise.

An envelope must name the axes it does **not** cover. Naming only the covered ones is exactly
what lets a half-blind envelope read as a handled case — and a partial remedy is more
dangerous than none, because none leaves the question open while a partial one looks like the
handled one and never gets re-examined.

## Tests

| arm | predicted | result |
|---|---|---|
| `a_clean_merged_read_reports_every_tier_and_no_fault` | red then green | matched |
| `a_malformed_line_comes_back_as_a_value` | red then green | matched |
| `a_tier_that_cannot_be_opened_is_a_fault_even_though_it_skipped_no_lines` | red then green | matched |
| `load_merged_still_returns_none_only_when_neither_tier_exists` | green both, control | matched |
| `no_graph_in_either_tier_says_so_and_never_renders_an_empty_working_set` | red then green | matched |
| `an_empty_working_set_names_its_scope_so_it_cannot_read_as_absent` | red then green | matched |
| `a_populated_working_set_still_lists_and_names_its_scope` | red then green | matched |
| `absent_and_empty_do_not_render_the_same_thing` | red then green | **see below** |
| `a_damaged_tier_is_reported_in_the_block_not_only_on_stderr` | red then green | matched, and the red was **observed** — see below |

### The two commits' reds are not the same strength, and the weaker one is named

The **first** commit's three new arms went red **by failing to compile**, because the function
they call did not exist yet. That proves a missing symbol, not a wrong behaviour, and the
control could not run at all while the crate would not build. Its behaviour evidence is the
separate probe quoted earlier. **A compile error is not a red arm; it is a reminder to go and
get one.**

The **second** commit's four arms compiled and **failed on behaviour**. That is the red worth
having.

### The differential arm's red was not what it looks like

`absent_and_empty_do_not_render_the_same_thing` is the only one of the nine that can catch
this defect. It asserts nothing about whether either output is correct — it asserts the two
are **not the same string**. When the defect is two states rendering identically, an assertion
about either one alone cannot see it, and every other arm can pass while it sits untouched.

Its red, though, was a panic rather than a comparison:

```
called `Result::unwrap()` on an `Err` value: no graph in either tier ...
```

It unwrapped the absent `Err` and died before comparing anything. **That red did not
demonstrate that absent and empty render identically. It demonstrated that the absent path
errored.**

The identical-screen claim rests on the two `signal/mod.rs` arms quoted at the top of this
description, which were read. An attempt to measure it directly at the `run_signals` surface
failed to compile against `SignalOutput`'s public shape twice and was abandoned rather than
pushed at. **That claim is read and quoted, not measured, and it is weaker for it.** From this
commit onward the arm does real differential work; its first red simply was not the evidence
it appears to be.

### The unreadable-tier arm can be defeated by its own environment

Running as root, or on a filesystem that ignores permission bits, a `0o000` file opens fine
and the arm would pass without testing anything. It checks that first and prints `VOID` and
returns rather than passing. Confirmed here to have taken the real path, not the VOID one, by
re-running it with `--nocapture` and seeing no `VOID` line. A test with a silent early return
is indistinguishable from one that passed.

## Verification

Full green window on the committed tree, tree locked before the first leg, re-read after the
last, and the commit asserted to **be** that tree:

| leg | result |
|---|---|
| `cargo test --all-targets --no-fail-fast` | rc 0 — **89 suites, 1,153 passed, 0 failed, 3 ignored** |
| tree lock | `a01c97af…` before, after, and as `HEAD^{tree}` |

1,153 is #169's 1,140 plus the four cross-tier arms, the four prerequisite arms and the five
envelope arms, less none. Each step of that chain is checkable without re-running anything.

**The new `SCOPE:` line does NOT break `session_start_layout_test`.** Its exact `l == whole`
matches on the PROJECTS, TASKS and MILESTONES headers all pass, because those header lines are
byte-identical and the clause is a line of its own. That was the open risk on the draft of this
description and it is now closed by measurement, not by argument.

**The first window did not finish.** It was killed by the tooling partway through strict clippy
while the host reported 0.99 GB free. That turned out to be a mismeasurement rather than a real
shortage — inside the VM there was 14.27 GB available, and most of what the host attributed to
the VM was reclaimable page cache. Re-run in the foreground, it completed untouched.

## Three things the suite forced that the design did not ask for

**1. `LAYOUT` gains the new kind, and its position is a decision.**
`every_pushed_kind_has_a_place_in_the_layout` failed with `kind "working-set-scope" has no place
in LAYOUT` — a guard doing its job. It is placed **before** the projects / tasks / milestones
rows, not after. The trimmer takes the bottom of a rank first, so a scope clause sitting after
the rows would be trimmed **while the rows survived**, and rows with no scope is the precise
defect the clause exists to close. An honesty clause has to outlive the thing it qualifies.

**2. A fifth arm, and its red was observed rather than asserted.**
`a_damaged_tier_is_reported_in_the_block_not_only_on_stderr` was written *after* the fix, so its
red could only be asserted from the code. That was not good enough, because this arm is the
evidence justifying the test change below. So the one scope push was removed, the arm was run
and **seen** to fail, and the push was restored and the arm seen to pass. Observed, both
directions.

**3. `session_start_failopen_on_malformed_trig` asserts something different now, and the path it
used to exercise is filed as its own item.**

That test asserted `handle()` returns `Err` on a malformed graph. The `Err` came from
`store::load_graphs(&paths)?` in the `queries.toml` **fallback**, which sits below an
`if any_signal { … return Ok(()); }` early return and only ran when *every* signal was silent.
A malformed graph used to leave every signal silent. This PR makes the working-set block never
silent, so the early return always fires.

**So that fallback is now unreachable.** That is a behaviour change outside this PR's scope and
it is **not resolved here**. It is filed as item **QTF-1** with three options and a
recommendation, for the operator to rule on — and it was filed *before* the test was touched,
because once the test stops asserting the old behaviour nothing else points at the dead path.

The test keeps its **name**. Its graph-unhealthy assertion is unchanged. `is_err` is **replaced,
not weakened**: the contract worth guarding was never an internal `Err` the dispatch wrapper
discards, but "a malformed graph does not pass silently" — now checked in the text session start
actually prints. What it used to assert, why it changed, and the item id are written at the test.

### A reader fault of mine, recorded in that test because it is worth more than the assertion

The first version of that new assertion read `out.fragments().parts()`, found no working-set
block, and failed. It was one line from being reported as *"the envelope does not fire here"*.
It fires fine. `push_signals` parks the signals in their own field and only merges them in
`finish()`, so the fragments alone **never** contain the block. The reader could not see the
state it existed to watch. Re-pointed at `finish().text`, and the wrong reader is recorded in a
comment at the assertion so the next person does not repeat it.


ChristopherKahler:
Landed on `main` as `848e434` and `621dfef` — cherry-picked rather than merged as a branch.

One follow-up was needed on integration: `main` carries a `deferred` field on `Section`, from the deferral lane, that this PR's two new `working-set-scope` sections did not set — 2 compile errors. Resolved as `deferred: 0` in `0bf407a`, with the reason written at both sites: a scope section states what was read, it does not list records, so it has no deferred ones of its own.

`main` is `2d3803babcc5db38e5517d205a03c62ecf31e835`. Closing: the content is on `main`.

PR #171, opened 2026-09-21
QTF-1: the operator queries.toml renders beside the signals

ChristopherKahler:
# QTF-1: the operator's `queries.toml` renders beside the signals

**Stacked on #170.** Base branch is `plover/0160-honesty-envelope`, so the diff here is one
commit, `75193fe`. GitHub retargets this to `main` as the stack merges.

## The defect

**An operator who wrote a `queries.toml` got nothing, and was told nothing.**

The ad-hoc query step sat below `if any_signal { … return Ok(()); }`, so it ran only when
**every** signal was silent — "fallback" in the strict sense of *instead-of*. On any workspace
with real data a signal always speaks, so the step never ran. The file parsed, the config
loaded it from both tiers, `LAYOUT` already reserved the block a place, and the code was
unreachable.

`docs/settings-hook-config.md` says session start *"Runs SPARQL queries from `queries.toml` (or
embedded defaults)"*. **A feature that silently does nothing looks exactly like a feature the
operator configured wrongly.**

## The fix

`render_adhoc_queries` is called **once**, before the signal early-return, for both paths.
Queries render beside the signals. Placement is decided by `LAYOUT`, which already carried
`("queries", Rank::Tail)`.

**This changes what session start prints for every operator who has a `queries.toml`.** That is
the point of the change, not a side effect — which is why it is its own commit and its own PR
rather than folded into #170.

### Fail-open, deliberately

The old path used `store::load_graphs(&paths)?`, a strict load that aborted the whole handler.
Now a graph that will not parse strictly costs the queries block and nothing else. The
unhealthy-graph warning is already collected by then, so a bad graph is reported by the block
that exists to say so — never by the silent absence of an unrelated one.

The `rdf`, `rdfs` and `xsd` prefixes carry across unchanged. Dropping them would have silently
broken any operator query that uses them, which is this commit's own defect pointed the other
way.

## How it was ruled, and a conflict of interest named

Filed by this lane as **QTF-1** after the honesty envelope (#170) made the last reachable case
unreachable: the envelope makes the working-set block never silent, so the early return always
fires. **Filed before the test that revealed it was touched**, so the record of the loss did not
depend on the thing that revealed it. Three options were recorded with a recommendation.

Ruled **option 2** by `finch` on Chris's delegated authority, 2026-09-21. **Chris can veto; the
ruling stands until he does.**

**Option 2 was this lane's own recommendation, so the ruling is NOT independent confirmation of
it and must not be read as any.** What `finch` added is what made it urgent rather than merely
right: **option 3, leave it silent, is what arrives if nobody decides**, and deciding-by-default
is the thing Chris ruled out. Option 1 would delete something the docs promise.

## Test, with an observed red

`an_operator_queries_toml_renders_beside_the_signals`

The fixture holds a project, so active-awareness **will** speak — a fixture with no signals
would have passed before the fix and proved nothing. The arm asserts the signal is present for
exactly that reason, then asserts the operator's own query rendered too.

**Red observed, not asserted.** The source change alone was stashed, the arm run and seen to
fail on *"a queries.toml must render even when signals speak"*, then restored and seen to pass.

## Verification

| leg | result |
|---|---|
| `cargo test --all-targets --no-fail-fast` | rc 0 — **89 suites, 1,154 passed, 0 failed, 3 ignored** |
| tree lock | `78c9c3f7…` before, after, and as `HEAD^{tree}` |

1,154 is #170's 1,153 plus this one arm.


ChristopherKahler:
Landed on `main` as `77c6839` — cherry-picked rather than merged as a branch.

**This change was correct, and it exposed a real contract break sitting behind it.** Rendering the operator's `queries.toml` blocks beside the signals meant the *shipped* queries began listing records that are supposed to be hidden: `recent-activity` selects anything carrying `lastActive`, and both deferred forks and archived reminders carry one.

Two tests caught it in sequence, the second hiding behind the first:

- `cold_handoffs_and_forks_are_deferred_in_the_tier_that_holds_them` — "a deferred fork is still listed at session start"
- `archived_never_shows_at_session_start_and_remove_still_deletes` — "an archived reminder reached session start"

Bisected to this commit by re-running each cherry-pick with the `Section` fix applied on top, because two of the four do not compile standalone. Fixed in `2d3803b`: `recent-activity` and `blocked-items` now filter `deferred` and `archived`. `active-projects` already required `status "active"`, so it could never match one.

`main` is `2d3803babcc5db38e5517d205a03c62ecf31e835`. Closing: the content is on `main`.

